Issue #9, opened 2018-03-29
Rationalise filenames


nicholst:
Hi @BryanGuillaume ,

As the use of SwE is growing, with @TomMaullin, we've been looking at doing several updates to SwE to improve usability. 

This is the beginning of a discussion on how to rationalise the filenames of output across different modes of the program.  Ideally I'd like to follow Anderson's general approach that he uses in [PALM](https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/PALM)-like naming system; see PALM's [Output files documentation](https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/PALM/UserGuide#Output_files).  However, that really only covers statistics and p-values and not everything else (frankly, I'm happy to keep with the cov_vis_* convention you use, but we do need to tidy up the various different ways of representing the original test statistics (T,F) with varying DF, and the converted Z,X images.

I had hoped to offer a concrete proposal but I am out of time right now and have instead just produced an inventory of the various filenames. 

Let me know if you have time/energy/interest to work with us on this.  No worries if not

-Tom

### An adaptation of PALM's conventions for our statistics and p-value outputs 

File name | File contents
-- | -- 
`swe_{unit}_{stat}_c{#}.{ext}` | Statistic for contrast `c{#}`.
`swe_{unit}_{stat}_{pval}_c{#}.{ext}` | P-value for  contrast `c{#}`.

Where we change to using a `swe` prefix to avoid confusion with SPM output, and...

Field | Description
-- | --
`{unit}` | The "unit" of type of inference; as with PALM, this can be `vox` for voxelwise, `vtx` for vertexwise, `dat` for any data that doesn't necessarily have a spatial meaning (e.g., data entered in mat format), `clustere` for cluster extent (allowing for, some day, `clusterm` for cluster mass support, and `tfce` for TFCE).
`{stat}` | The name of the statistic, which can be `Tstat` for the t-statistic, `Fstat` for the F-statistic, `Zstat` for the Z-statistic, `Xstat` for the chi-squared-statistic; if the statistic is based on an inversion (probability integral transform) of Wild Bootstrap p-values, then `WB` is appended. 
`{pval}` | The type of p-value stored in the image. It can be `uncp` for uncorrected permutation p-values, `FWEp` for FWER-corrected p-values, `FDRp` for  FDR-adjusted p-values 
`c{#}` | Contrast index, in the same order as in the files supplied with the -t and/or -f options.
[...]

PR #12, opened 2018-05-08
Output format and restructure

TomMaullin:
This PR is a cleanup of the current SwE-Toolbox code. It addresses issues [8](https://github.com/BryanGuillaume/SwE-toolbox/issues/8) and [9](https://github.com/BryanGuillaume/SwE-toolbox/issues/9) on the original SwE-Toolbox repository by ensuring the correct files are output at the end of every run.

In addition several parts of `swe_cp_WB.m` have been restructured as, previously, many sections of code were repeated several times. Doing this allows easier and quicker code maintenance in the future and also opens up the possibility of unit testing which only tests specific functions/features of the code. This restructuring has been done very carefully and a wide range of use cases have been checked throughout to ensure that the output of the code has not been changed.


TomMaullin:
The maps created by the toolbox in this update have been verified to make sure they are exactly the same as the maps generated before the update. I have checked that all of maps in the use cases of surface and volumetric, T and F tests, are unchanged when the seed is the same. I have uploaded the new and old outputs to google drive [here](https://drive.google.com/drive/folders/1VuTowuWnkaKD47ywz5l4jqenB0QuRxBB?usp=sharing) for comparison (note the names of these files have also changed - to help I've also included a text file listing the new and old filenames, see[ `Changed Names.txt`](https://drive.google.com/file/d/1IC2Z3Pop22I-JYjkl8wppJdjbVTYPHP5/view?usp=sharing)).

This PR is now ready for review!

nicholst:
@TomMaullin - Really great work on this PR!  A huge effort!!

I'm almost through reviewing this.  Much of it I can easily see is correct (i.e. replacing much-repeated code with functions) but sometimes something is subtle and I can't instantly see it's the right thing.

One of these is the movement of the assignment of `cCovBv` up a block, instead of a bunch of conditionals; in particular I'm looking at L767 of your branch's `swe_cp_WB.m` (see the [diff](https://github.com/NISOx-BDI/SwE-toolbox/pull/12/files#diff-8464b4a3267760826308f04b9ab0175aR767))... can you just talk me through this particular change?

TomMaullin:
```
@TomMaullin - Really great work on this PR! A huge effort!!

I'm almost through reviewing this. Much of it I can easily see is correct (i.e. replacing much-repeated code with functions) but sometimes something is subtle and I can't instantly see it's the right thing.

One of these is the movement of the assignment of cCovBv up a block, instead of a bunch of conditionals; in particular I'm looking at L767 of your branch's swe_cp_WB.m (see the diff)... can you just talk me through this particular change?
```

Hi @nicholst , 

Yes, sure! So in the original code there was an if statement on [line 931](https://github.com/NISOx-BDI/SwE-toolbox/blob/master/swe_cp_WB.m#L931) which was essentially doing the following:

```
if isfield(SwE.type,'modified')
     % Assign a load of variables needed to calculate cCovBc in this case but not cCovBc itself.
else
   % Assign cCovBc
end
````

(Where the else can be found on [line 976](https://github.com/NISOx-BDI/SwE-toolbox/blob/master/swe_cp_WB.m#L976) ).

This was then immediately followed by an if statement which essentially did the following:

```
if (nSizeCon == 1) (see [line 985](https://github.com/NISOx-BDI/SwE-toolbox/blob/master/swe_cp_WB.m#L985))
     if isfield(SwE.type,'modified')
          cCovBc = weightR * Cov_vis;
     end
     % Calculate score and do a T stat hypothesis test if needed for clusterwise
else  (see [line 1062](https://github.com/NISOx-BDI/SwE-toolbox/blob/master/swe_cp_WB.m#L1062) )
     if isfield(SwE.type,'modified')
          cCovBc = weightR * Cov_vis;
     end
     % Calculate score and do a F stat hypothesis test if needed for clusterwise     
end
```

I thought it would be sensible to move the assigning cCovBc as it is repeated unnecessarily this way and if any code was ever added between these blocks (i.e. between lines 976 and 985) this could cause errors as cCovBc has been assigned in one case and not the other. (Note: around 200 lines had been shortened before this edit which is why these changes now appear at line 767 as you mentioned in your comment)

**Comment by TomMaullin** [comment]
```
Another question:

It seems that the maximum score is no longer saved... is that the case? Did we agree on that?

I'm looking at L2262 in the original (master) version of swe_cp_WB.m (see diff)
```

I have removed this, yes. Apologies if I was mistaken - I thought we said we were going to remove all maps bar the output maps for viewing. I can happily add this back in though if you need!

**Comment by TomMaullin** [comment]
```
Another random thing: Any idea why there are * 1 expressions in the code? Like at L1440 in your branch of swe_cp_WB.m? At first I thought it was about managing types (i.e. logical vs real) but I couldn't create a case where that was a problem.
```

I'm glad you have said this because I honestly have no idea. I did wonder this myself and my thoughts were it must be type related as well but also couldn't create a problem. One thought I did have is that maybe it's for backwards compatibility - perhaps it causes errors in older versions of matlab. I wasn't sure why this had been added so avoided touching it but if you agree I think these should be removed.

TomMaullin:
Hi @nicholst ,

Apologies you were right the min and max score variables were not previously saved in the SwE structure. I have now added this in! (See [lines 1599-1611](https://github.com/NISOx-BDI/SwE-toolbox/pull/12/commits/3c7fa07c234a7a924dba386e26a013b72e645239#diff-8464b4a3267760826308f04b9ab0175aR1599) of `swe_cp_WB.m`). Please let me know if you have any further feedback on the PR or have any more requested changes to be included here!

nicholst:
OK @TomMaullin !  I think we're ready to go *after* you do the following

- [ ] Update `swe_getSPM.m` to roll-back the changes that make it seem like WB works with SPM results display.
- [ ] Edit `swe_contrasts.m` and `swe_cp.m` to consistently use `swe_create_vol.m`; these are two functions that have the repetitive code that this function eliminates.

Note that one change I've added to my PR on your branch (https://github.com/TomMaullin/SwE-toolbox/pull/8) is that `swe_create_vol.m` now consistently uses `float64` instead of `float32`... while this does double the disk usage, it is consistent with SPM (in spm_spm.m, all but the beta images are float64; the contrasts like the beta's are float32...not sure why, but don't see why to be stingy on precision, especially considering internal precision is float64)

THEN, once you're done, ping me, I'll take quick look and then if all looks good we'll merge.

TomMaullin:
Hi @nicholst ,

Your changes looked good to me! I have run the image deleting code to check I understand how this is working and can verify it works well and makes sense.  I have merged your branch into this one and I believe I have addressed everything you mentioned above! Please let me know if you have any further feedback, else this looks good to merge!

nicholst:
I'll merge this now... but before do much more, can we finish rationalising the filenames, as in https://github.com/NISOx-BDI/SwE-toolbox/issues/9#issuecomment-388373867 in a quick follow up PR?  (I.e. assume changing the output names will be a pain if you have to update tests)

PR #13, opened 2018-05-08
Testing

TomMaullin:
This PR implements rigorous unit testing for the SwE toolbox.


TomMaullin:
Update:

The tests I am implementing for the SwE-toolbox will have to first run the SwE-toolbox itself. After much testing I now believe that the SwE-toolbox cannot be run in Octave and there are no quick fixes to the number of errors it encounters due to functions it relies on that are currently implemented in MatLab but not in Octave. This isn't a complete surprise but because of this I do not believe tests can currently be done using a Travis testing suite with a docker container running Octave. 

The tests I will implement in this PR will be designed to run locally only. Maybe with the next Octave release Travis testing might become more feasible.

TomMaullin:
Hi @nicholst ,

Apologies this has taken a little while longer than I had expected but this branch now contains a rigorous and heavily checked testing suite!

There are 8 tests, all with names of the form `a_b_c` where `a`, `b` and `c` all come from the below list:

- `wb`  - Wild Bootstrap
- `p`     - Parametric
- `t`      - t test
- `f`      - f test
- `img` - volumetric `.img` and `.hdr` input
- `mat` - surface `.mat` input

As the 4 tests which perform wild bootstraps each can take up to 25 minutes I have implemented the tests so that they run in parallel using Travis-CI's new parallel testing features (an example log of which can be found [here](https://travis-ci.com/TomMaullin/SwE-toolbox/builds/75728152)) . The result of this is that the entire suite usually runs in under 30 minutes.

Each test is verified against a ground truth dataset, the repository for which can be found [here](https://github.com/TomMaullin/SwE-toolbox-testdata), and the testing is done using octave via @cmaumet 's SPM docker container `cmaumet/octave-spm` (which can be found [here](https://hub.docker.com/r/cmaumet/octave-spm/)).

To see some previous runs of the tests and more previous Travis logs I have my own version of this PR which can be found [here](https://github.com/TomMaullin/SwE-toolbox/pull/11). To activate the tests on  this repository, activate the Travis-CI. I have sent you an email requesting this.

nicholst:
Hi @TomMaullin, this looks all good.  The only thing, for sake of tidiness, is all the conditionals needed for `spm_progress_bar`; since Octave is the ultimate global, wouldn't be better to create `snpm_progress_bar`, which does nothing but check if Octave is going, and if not then call `spm_progress_bar`?  

*OR* is this a hacky workaround specific to testing, and really, it should be
```
if ~A_Test_Is_Running
  spm_progress_bar...
end
```
?

TomMaullin:
Hi @nicholst , 

Yes, I agree, this was my main concern as well. I think the correct behaviour would probably be the `if ~A_Test_Is_Running` suggestion you have made as the loading bar does work in octave but produces ascii art which spams the test logs. 

This would mean passing some form of `A_Test_Is_Running` field into either the SwE object or directly into the functions as a separate object though. Does this seem okay to you? If so I'll change this now.

swe_progress_bar.m
```diff
@@ -0,0 +1,17 @@
...
+% spm_progress_bar displaying in octave. The progress bar can be 
+% used in octave but it creates ascii art which spams the test
+% logs. 
+%
+% Author: Tom Maullin (08/06/2018)
+%
+% ===============================================================
+function swe_progress_bar(varargin)
+
+	if ~isOctave
```
nicholst: Just clean up the indentation here please.

swe_progress_bar.m
```diff
@@ -0,0 +1,17 @@
...
+% spm_progress_bar displaying in octave. The progress bar can be 
+% used in octave but it creates ascii art which spams the test
+% logs. 
+%
+% Author: Tom Maullin (08/06/2018)
+%
+% ===============================================================
+function swe_progress_bar(varargin)
+
+	if ~exist('OCTAVE_VERSION','builtin')
```
nicholst: Just tidy up the indentation here... looks like you have a tab where you think you have a space

swe_getSPM.m
```diff
@@ -187,7 +187,9 @@
 %-GUI setup
 %--------------------------------------------------------------------------
 %spm_help('!ContextHelp',mfilename)
-spm('Pointer','Arrow')
+if ~exist('OCTAVE_VERSION','builtin')
```
nicholst: For my edification, what gets screwed up in Octave with `spm('pointer','Arrow')`?

nicholst:
Some tiny comments embedded in the code... I'll be getting busier today... it's ready to merge when you're happy.

swe_getSPM.m
TomMaullin: I think this must have been a mistake...

TomMaullin:
Okay great! All comments are addressed - I'll now merge!

PR #19, opened 2018-05-30
[WIP] File name change

TomMaullin:
This PR is a work in progress. This PR contains all filename changes discussed in issue #9.


TomMaullin:
Hi @nicholst ,

This PR is now ready for review! The only filename we hadn't previously included was `con_{#}` which was the Beta values for each contrast in the parametric analyses. I have renamed these to `swe_vox_beta_c{c#}` and added this change to your comment on PR #9 . 

Please let me know if you have any further feedback!

TomMaullin:
Hi @nicholst ,

When we discussed in person we decided to go with `zTstat` instead of `Zstat` which is the reason behind the `lower(eSTAT)` you have mentioned above. Would you still like to go with this convention?

The other changes make sense to me! I'll implement these!

nicholst:
Sorry, I had forgotten about 'zTstat'... no, that's still a good idea :) let's keep it.

TomMaullin:
Hi @nicholst ,

I have now addressed all of your feedback above! Let me know if you have any further feedback and if not we can merge!

swe_contrasts.m
```diff
@@ -114,13 +123,13 @@
...
               %------------------------------------------------------
-              xCon(ic).Vcon = sprintf('con_%04d.mat',ic);
+              xCon(ic).Vcon = sprintf('swe_vox_beta_c%02d%s',ic,file_ext);
               save(xCon(ic).Vcon, 'cBeta')
             else
               %-Prepare handle for contrast image
               %------------------------------------------------------
               xCon(ic).Vcon = struct(...
-                'fname',  sprintf('con_%04d.img',ic),...
+                'fname',  sprintf('swe_vox_beta_c%02d',ic,file_ext),...
```
nicholst: Missed a `%s` here...

swe_cp.m
```diff
@@ -489,8 +490,8 @@ function swe_cp(SwE)
     %-Initialise standardised residual images
     %----------------------------------------------------------------------
     % for i = 1:nSres
-    %     VResI(i) = swe_create_vol(sprintf('ResI_%04d.img', i),...
-    %                               DIM, M, sprintf('spm_spm:ResI (%04d)', i),...
+    %     VResI(i) = swe_create_vol(sprintf('ResI_%02d.img', i),...
+    %                               DIM, M, sprintf('spm_spm:ResI (%02d)', i),...
```
nicholst: Sorry to have missed this before... as per https://github.com/NISOx-BDI/SwE-toolbox/issues/9#issuecomment-388373867 this should be `swe_vox_resid_y%04d`

swe_cp.m
```diff
@@ -1088,30 +1089,30 @@ function swe_cp(SwE)
     fprintf('%s%30s',repmat(sprintf('\b'),1,30),'...saving results'); %-#
 
     mask = Cm;       
-    save('mask.mat', 'mask');
+    save(sprintf('swe_vox_mask%s',file_ext), 'mask');
```
nicholst: Just to double check: `file_ext` is `.mat` and not `mat`, right?

swe_cp_WB.m
```diff
@@ -499,7 +498,7 @@ function swe_cp_WB(SwE)
       else
         descrip = sprintf('adjusted unrestricted residuals (%04d)', i);
       end
-      VResWB(i) = swe_create_vol(sprintf('ResWB_%04d.img', i), DIM, M, descrip);
+      VResWB(i) = swe_create_vol(sprintf('ResWB_%04d%s', i, file_ext), DIM, M, descrip);
```
nicholst: I didn't specifically consider this case... but maybe `swe_vox_resid_y{y#}` is OK here? (i.e. don't need to specifically note these are WB residuals... unless both WB and non-WB residuals co-exist)

swe_cp_WB.m
```diff
@@ -513,52 +512,44 @@ function swe_cp_WB(SwE)
       else
          descrip = sprintf('unrestricted fitted data (%04d)', i);
       end
-      VYWB(i) = swe_create_vol(sprintf('YfittedWB_%04d.img',i), DIM, M, descrip);
+      VYWB(i) = swe_create_vol(sprintf('YfittedWB_%04d%s',i,file_ext), DIM, M, descrip);
```
nicholst: Missed this one too in my table of revised names; let's go for `swe_vox_fit_y{y#}`

nicholst:
OK!  Done with my reviewing... just some quick fixes needed.

swe_cp.m
TomMaullin: I can confirm this is '.mat'!

TomMaullin:
Okay, I think that's all feedback addressed! Let me know if you have any further feedback!

PR #22, opened 2018-06-08
WB display

TomMaullin:
This PR implements the display feature for wild bootstrap analyses.

PR #23, opened 2018-07-05
Updated Results GUI

TomMaullin:
This refactors the results options to match that of SnPM. This follows directly from PR #22 and should only be merged once PR #22 has been reviewed and merged.


TomMaullin:
Hi @nicholst ,

All tests pass and this PR is now review ready (once PR #22 has been merged that is)! I believe I have ensured that all of the display GUI options are now correct/match what we discussed yesterday but it may be worth playing around with the results option of the GUI just in case I have misunderstood anything (there were a lot of use-cases to account for). Or if you want I could send you some example screenshots of use-cases you wish to double-check.

Once PR #22 is merged I will rebase this branch onto the master branch to make it clearer what the changes I have implemented here are.

nicholst:
When using "Results" (any inference option) I get a message printed to the command window just as MIP/Results window is populated:

```
ans = 

  UIControl (hReg) with properties:

              Style: 'frame'
             String: ''
    BackgroundColor: [0.5800 0.7700 0.5700]
           Callback: ''
              Value: 0
           Position: [1.2111 1.2111 484.4444 230.1111]
              Units: 'pixels'

  Use GET to show all properties
```

nicholst:
Aside from previous comment, looks great!

nicholst:
One improvement:

When doing Results -> Clusterwise -> FWE, there is no choice of threshold as it is pre determined, but then the GUI has (as in SPM) `& extent threshold {FWE P}`, where the "&" is alluding to a previous query (`threshold {F or p value}`) that is missing since the threshold is predetermined.

To help the user, let's issue a place-holder message that shows the cluster-forming threshold but makes it clear it cannot be changed.  I believe the 'display' functionality of `spm_input` can do it, like
```matlab
spm_input(`threshold {F or p value}`,'+1','d','(pre-set: X=X.xxx P=X.xxx)')
```
where X would be replaced by Z as appropriate, and X.xxx would be the corresponding actual values.

TomMaullin:
Hi @nicholst ,

I have now addressed the issue of the `spm_input` display on this branch and the printout error reported has been addressed in PR #22 and should disappear when merged. I have also addressed issue #28 (Display "whole brain" p-values right away) here as it seemed relevant to this PR. Please let me know if you are happy to merge/have any further feedback!

nicholst:
@TomMaullin - Ack - Forgot to ask about something!  The message in the GUI display showing the "already set" threshold is great, but it showed an F-threshold and a P-value... is it unambiguous what this F-threshold is given the issue of local DF and possibly using an X2 distribution?

Maybe, is it just safer to record it as an uncorrected P-value?

How is the cluster forming threshold stored internally?  P-value or F-value?  And if F, on exactly what image is applied to?

nicholst:
How can the user enter it? If they can *only* enter it as a P-value, show
it only as a P.  If they can enter it as either X or P, then keep it as is
but, indeed, show it as a X instead of F (if latex interpreter is supported
in that context, try \chi^2_1)

Issue #16, opened 2018-05-21
Add warning when intercept is not in the model


nicholst:
While not strictly a "bug", a model with no intercept is such a catastrophic modelling decision it basically is one.

@TomMaullin - Let's discuss how to add this ASAP.

nicholst:
A simple check would be 
```matlab
all(one(N,1)==P*ones(N,1))
```
but probably should be done as an `abs(diff)>sqrt(eps)` to allow imperfect numerical precision.

TomMaullin:
This issue has been addressed by PR #30

PR #30, opened 2018-07-10
Intercept warning

TomMaullin:
Hi @nicholst ,

This is a quick PR that addresses two issues:

1) #16 - An intercept not included in the model now throws a warning.
2) In the previous PR ( #23 ) we realized the pre-set WB cluster forming threshold was given as a P and F threshold. As the user was never able to enter the threshold as a stat this has been changed to just P.

Please let me know if you have any feedback!

### DIFF
--- a/swe_getSPM.m
+++ b/swe_getSPM.m
@@ -947,8 +947,8 @@ if ~isMat
                   
                   % We should display the cluster forming threshold that
                   % was used.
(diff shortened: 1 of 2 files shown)


TomMaullin:
The travis bug has been fixed (there was already a variable named P!). Now merging!

nicholst:
> Sorry! One change:

Can we upgrade the warning to an error? On reflection there really is never any case where there can be no intercept.

Issue #24, opened 2018-07-06
Make an informative footer


TomMaullin:
Currently there is no footer information in the display window. This should be addressed in future PRs.

nicholst:
Here is SPM's footer: <img width="681" alt="screen shot 2018-07-08 at 14 35 11" src="https://user-images.githubusercontent.com/5155907/42453115-a6f5f87e-8383-11e8-8dcc-c282a2c5b1e8.png">

TomMaullin:
Hi @nicholst ,

I see, it's difficult for me to tell which variable is which in the SwE object for some of these so below is what I think should go in each of these fields - does anything here look obviously incorrect to you?

> Number of subjects: X1 X2 ... 

This should be `SwE.Gr.nSubj_g`.

> Number of visits ([Mn Max]): [Mn1 Mx1] [Mn2 Mx2] ....

This should be `SwE.Vis.nVis_g`. I'm not sure what Mn1 and Mx1 would be here? Is this suggesting individual subjects within a group can have a different number of visits? Does the current code support this?

> Number of predictors: P

I think this would be `size(SwE.xX.X, 2)`.

> Contrast DF: P1

I think this would be `rank(SwE.xCon(Ic).c)`.

> Error DF: (Type X): (min) X.X, (median) X.X, (max) X

I think this would be done by:

```
if dof_type ~= 1 
      
    edfmin = min(vedf)
    edfmax = max(vedf)
    edfmed = med(vedf)

else 

    edf = vedf(1)

end
```

Where `vedf` is just the array of edf values for each voxel.

Does this all sound correct to you?

nicholst:
> Number of subjects: X1 X2 ...
> This should be SwE.Gr.nSubj_g.

Yup!

> Number of visits ([Mn Max]): [Mn1 Mx1] [Mn2 Mx2] ....

There is a visit indicator somewhere.... you’ll have to do a loop over subjects in each group to find min and max #’s. 

(there is no assumption of equal visits per subject… This is crucial… It is a key strength of SWE.)

> Number of predictors: P
> I think this would be size(SwE.xX.X, 2).

Yes!

> Contrast DF: P1
> I think this would be rank(SwE.xCon(Ic).c).

Yes!

And yes on error DF!

-Tom

nicholst:
Please look at TabDat! I forgot about this. There is a field for footer information. When the user right-clicks you can see it. Check if you can find out about this as well.

[...]

PR #34, opened 2018-07-26
Footer creation

TomMaullin:
It aims to address issue #24 by providing an informative footer on the display for all analyses with img/nifti input.


TomMaullin:
Hi @nicholst ,

This footer is now complete, all tests pass and this PR is review ready!

Below are some screenshots of the updated display with footer in 4 use cases! Please let me know if you have any thoughts/feedback!

Case 1: F Contrast, Wild Bootstrap

image

Case 2: F Contrast, Parametric

image

Case 3: T Contrast, Wild Bootstrap

image

Case 4: T Contrast, Parametric

image

nicholst:
Great!

A couple of things: 

The page numbers are really important. If the peaks need more than a page then you need those number/buttons to get to the other pages. 

Number of subjects & visits can go on one line. 

Number of predictors & contrast DF can also go on one line. 

*When* the error DF are of a type that is *not* spatially varying, then we don't want the min/med/max. That should only appear when it is *possible* for error DF varies. 

Finally, for the items that also are on the usual SPM footer , can they appear in the same relative location? This means the voxel information is on the bottom of right column. Info on number of subjects and visits is SwE-specific but could go where there is space, bottom of left column. 

Let me know if all of this makes sense.

TomMaullin:
Hi @nicholst ,

> The page numbers are really important. If the peaks need more than a page then you need those number/buttons to get to the other pages.

Sorry - my commit messages probably are a bit misleading here. In this display the page number was given twice - once in the bottom center where it says `Page x` and once at the bottom right where it says `x\y` and gives a button to move from page to page. I removed the `Page x` line as the footer was now so large it wrote over it and the text was unreadable. I tried making the text in the footer smaller but it had to become very small and wasn't very legible. 

I thought this would be okay as the page number is still displayed alongside the button to move from page to page in the bottom right hand corner (see Case 2 in my previous comment for example). 

The buttons to move from page to page are still present and in the expected location and the page number is still visible. Apologies if this is an issue - I can try and add it back in after making the below edits if you'd like? 

> Number of subjects & visits can go on one line.
   Number of predictors & contrast DF can also go on one line.

Ah okay! Yes This should clear some space!

> *When* the error DF are of a type that is *not* spatially varying, then we don't want the min/med/max. That should only appear when it is *possible* for error DF varies.

Ah apologies! This has been implemented but from #24 I thought that `Type 1` degrees of freedom were the only spatially varying degrees. Should there be some sort of additional check for `Type 2` and `Type 3`?

> Finally, for the items that also are on the usual SPM footer , can they appear in the same relative location? This means the voxel information is on the bottom of right column. Info on number of subjects and visits is SwE-specific but could go where there is space, bottom of left column.

Ah okay so just to confirm should the order be:

```
First Column:

Height threshold
Extent threshold 
vox T/F(5% FWE), vox P(5% FDR), clus k(5% FWE)
Number of subjects, Number of visits
Number of bootstraps

Second Column:

Error DF
Contrast DF, Number of predictors
Volume
Voxel Size

```

nicholst:
Space saving idea:  Combine Height & extent thresholds:
```
Threshold: Height Z = X.XX, p = X.XX; Extent k = X.XX vox. 
```
Does that work?

> Ah apologies! This has been implemented but from #24 I thought that Type 1 degrees of freedom were the only spatially varying degrees. Should there be some sort of additional check for Type 2 and Type 3?

I've muddled the waters here I'm afraid, and realise there's one more thing to report!

In Bryan's conventions, "Type" can refer to small sample corrections applied to to the residuals before computing covariance _or_ Type can refer to Degrees of Freedom.  

Residual adjustment Types are 0/1/2/3/C2/C3.  From the help, it is clear that one should *only* ever use Type C2.  But, still, if there is space, it would be good to report: `Resid. Adj.: Type C2` (or whatever has been selected).

For Degrees of Freedom it is *Approximation* type, there is naive/approx I/approx II.  Naive is *not* spatially varying; both approx I and II are spatially varying.  (Approx II is the default with modified SwE, and Approx I is recommended if there is no missing data; but if classic SwE is used only Approx I is used).

So!  DF should read: `Error DF (<type>): ...` where <type> is one of naive/approx I/approx II.

Does this all make sense?

TomMaullin:
Hi @nicholst ,

Ah I see, yes this is clear. The only thing I am unsure on is what about Approx III for dof? Is this spatially varying?

TomMaullin:
Hi @nicholst ,

Assuming dof type `Approx III` is spatially varying I have addressed all above feedback and all tests pass! Here are some screenshots to briefly show the updated display:

Case 1: T contrast, Wild bootstrap

image

Case 2: T contrast, Parametric

image

Please let me know if you have any further feedback/suggestions!

nicholst:
I swear Type III was spatially varying, but maybe not in this balanced case here.

Anyway, almost there:  Let's just check if the DF are all the same... if so, just list one number.  
If they are different, then give three numbers but format them like number of visits, giving the meaning first, like:

([Min Med Max]) [9.88 11.1 12.3]

OK?

Make that change and then we can merge!

TomMaullin:
This has now been done. The code now checks whether `abs(edf_min-edf_max) < 1^-10` and if it is, the footer only reports one value, otherwise it reports the minimum median and maximum. Here is an updated image of Case 1 from above:

image

Once the tests pass would it be okay for me to merge?

nicholst:
Awesome! Yes, merge when ready.

Issue #25, opened 2018-07-07
Allow multiple regressors to be specified at once


nicholst:
When a design matrix is created and ready, it is very cumbersome to enter it column-by-column.

Should have an option to enter multiple columns at once, and then enter an array of names, one per column.

nicholst:
Grab example code from https://github.com/spm/spm12/blob/master/config/spm_cfg_factorial_design.m#L744-L777 ... adding a "Multiple Covariates" option.

Additionally, add an item for "Names" so that, e.g., if a text file is supplied, the names can be supplied.

TomMaullin:
Hi @nicholst ,

I was just wondering if you have any example `.mat`/`.txt` files I could use look at as an example of the code I should add?

nicholst:
Any plain text file that can be parsed by Matlab into a matrix with nScan rows and any number of columns.  I.e. anything that works as 
```
X = load('Vars.dat');
```
where, e.g. you could generate `Vars.dat` like:
```
nScan=30;
nVar = 5;
Vars = rand(nScan,nVar);
Nms = strcat('Var_',num2str([1:nVar]'));
save Vars.dat Vars -ascii 
```

Now sure how you are handling the names... they could come from a plain text file or you could require them in a cell array like `Nms`.

TomMaullin:
Ah I see! Thank you... so I should assume they are all in one matrix? I originally thought we were thinking about data generated in a way something along the lines of:

```
Vars = struct('Var1', [1 1 1], 'Var2', [0 1 2])

save('Vars.mat', 'Vars')
```
(supposing there were 3 subjects).

The only reason I thought this was that each variable could then have it's associated name (`Var1`, `Var2` etc) stored alongside it.

TomMaullin:
This issue has now been addressed by PR #45

PR #44, opened 2018-08-31
[WIP] Adding multiple covariate options

TomMaullin:
This PR is a work in progress. It aims to address issue #25 by adding options to specify multiple regressors at once, either using a file containing a structure or two files, one containing a matrix and one containing a list of filenames.

### DIFF
--- a/swe_cfg_design.m
+++ b/swe_cfg_design.m
@@ -137,29 +137,89 @@ iCC.labels = {
 iCC.values = {1 2 3 4 5 6 7 8};
 iCC.val    = {5};
 % ---------------------------------------------------------------------
-% cov Covariate
+% Individual Covariate
 % ---------------------------------------------------------------------
-cov         = cfg_branch;
-cov.tag     = 'cov';
-cov.name    = 'Covariate';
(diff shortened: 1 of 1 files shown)


nicholst:
Current help in SPM for multiple covariates:

> Select the *.mat/*.txt file(s) containing details of your multiple covariates. 
> 'You will first need to create a *.mat file containing a matrix R or a *.txt file containing the covariates. Each column of R will contain a different covariate. Unless the covariates names are given in a cell array called ''names'' in the MAT-file containing variable R, the covariates will be named R1, R2, R3, ..etc.'

Ours will read like:
> Select the *.mat/*.txt file(s) containing details of your multiple covariates. 
> 'A *.mat file must contain a matrix R, and may optionally contain a cell array called 'names' with covariate names. A *.txt file can be any file that can be read by Matlab's IMPORTDATA function, which includes a data-only text file, with spaces, tabs or comma separators.  A header row is allowed, and if present will be used to obtain the covariate names.  If names for the covariates are not give they will be named R1, R2, R3, ..etc.'

Issue #29, opened 2018-07-09
Remove fig-based GUI window - Adopt Batch-only system


nicholst:
I like having a main 'GUI' for SwE (miss it in SnPM), but ultimately we want to be fully grounded in the Batch System so as to facilitate scripting.

However, even if we keep a GUI (that just opens up batch jobs), we  should get rid of this `swe_ui_main.fig` ... it is only editable via Matlab GUI editor, GUIDE.  While font sizes are tweaked in  `swe_ui_main.m` I'd rather just have a figure created from scratch each time.  Also, I have done some tests and found that the GUI based on `swe_ui_main.fig` is *very* slow on first click; I thought this was due to delays with the batch system firing up, but I don't think this the explaination.

TomMaullin:
Hi @nicholst ,

Just to clarify, if keeping the main GUI, is there a preferred editor you would like to build it from? Or would you prefer the GUI be built from scratch through the `uicontrol` functions?

nicholst:
Build from scratch.  Sounds like a pain, but we have templates (SPM or
SnPM) that we can clone.

TomMaullin:
Ah okay! No that's fine! Did you have any in particular SPM/SnPM menus in mind which you wanted the GUI to resemble?

TomMaullin:
Currently, the options are 'Specify Model', 'Run Model' and 'Results'.

nicholst:
Oh, right... then those are the 3 buttons we need!  I just realised that the oldest SnPM on GitHub has now GUI window.

Please look here: https://gist.github.com/nicholst/4eb26c999fd736b8ebc063fb4543272f

Around Line 92 is the start of the creation of the window.

TomMaullin:
This issue has been resolved by PR #46

PR #43, opened 2018-08-28
Comment Review

TomMaullin:
This PR addresses issue #14 by ensuring all comments have a similar format and do not include details of any functionality not currently included in the SwE code.

PR #46, opened 2018-09-05
GUI update

TomMaullin:
This PR updates the SwE GUI and addresses issue #29 by creating an interface similar to that of SnPM's.


TomMaullin:
Hi @nicholst ,

I have now made a new GUI based on that of SnPM's which looks something like the below:

image

I have also added the following features:

 - The 'Documentation' button which opens the NISOx website for SwE (The main reason I have added this was for the layout).
 - Two '?' buttons which open the comment sections added in PR #43 for reference. This minics the behavior of the previous SnPM menus.
 - The version number is also now displayed in the menu window for the users reference.

Please let me know what you think of these modifications and/or if you want anything to be changed/removed.

I am still waiting on the tests to run (the previous two runs failed because I stopped them manually myself) but please if you have a moment let me know what you think of the menu design and if possible whether it runs well on your mac as well.

Also included in this PR is the change to `swe_design` we have been discussing via email.

nicholst:
Looks good.  Only comment is totally superficial... the colors are garish!  

Can we inherit the SnPM colors?  Having the color set as a call to swe also seems sensible:
```matlab
elseif strcmp(lower(Action),lower('Colour'))
%=======================================================================
% snpm('Colour')
%-----------------------------------------------------------------------
% %-Developmental livery
% varargout = {[0.7,1.0,0.7], 'Lime Green'};
%-Distribution livery
varargout = {[0.8 0.8 1.0], 'Diluted Blackcurrent Purple'};
```
where we set `'BackgroundColor',swe('Colour')`

TomMaullin:
Hi @nicholst ,

I have entered the colourscheme found in the SnPM code you sent and this is the resultant interface. I just thought I had best confirm, is this the colourscheme you' had in mind?

image

TomMaullin:
Hi @nicholst ,

As all tests pass, so long as the above colour scheme is what you wanted, this PR is now ready for review! Please let me know what you think of the new GUI and batch options. I had to rename a few files to try and make the code layout consistent also - please let me know if you have any thoughts/suggestions/improvements and if you have a moment to try it out, whether everything runs correctly on Mac! Just to sum up what has changed in this PR:

 - The new interface has been added, which now includes:
   - The 'Documentation' button which opens the NISOx website for SwE in the matlab browser.
   - Two '?' buttons which open the comment sections added in PR #43 for reference.
   - The version number displayed in the menu window.
 - All SwE inputs now run through the batch window (with the exception of part of the `results display` section which we discussed has to be dynamic).
 - SwE can now be run as a toolbox through the SPM toolbox menu.

nicholst:
Fantastic!!  This is what I always wanted!

Just one issue: When I click 'Run Model'  the Batch system asks me for *both* the directory and the SwE file, but the Directory has already been specified.  Can you eliminate that?

BTW, it looks different for me on a Mac (see below) but I'm really not fussed (don't worry about it).

Fix the double-entry of directory and this can get merged!

screen shot 2018-09-06 at 17 02 05

Issue #21, opened 2018-06-08
Add small volume correction option


nicholst:
Would only be available for parametric, FDR corrections.

nicholst:
Only a priority if could be easily accommodated through code inherited from `spm_getSPM.m`

TomMaullin:
Hi @nicholst,

I have looked into this and found that to do this the following parameters are needed (at least):

```
 - xSwE.VRpv - RESELS per voxel (density) volume
 - xSwE.FWHM - Full width half maximum
 - xSwE.df - degrees of freedom 
```

I guess `df` could be set to the nominal degrees of freedom and we could ask the user to input the `FWHM` they used perhaps? I'm unsure how `VRpv` is calculated however.

nicholst:
Hi @TomMaullin 

Confused... we have a working assumption that SwE doesn't work with RFT.  As noted above, this is *just* for FDR corrections, since spm_P_FDR is quick to run.

Does this help?

TomMaullin:
Ah right apologies, yes. The code from `spm_results_ui.m` calls to `spm_VOI.m` for small volume corrections (see [here](https://github.com/spm/spm12/blob/master/spm_results_ui.m#L604)). This code uses:

- `VRpv` in the call to `spm_list` at line [215](https://github.com/spm/spm12/blob/master/spm_VOI.m#L215)
- `df` in lines [297-299](https://github.com/spm/spm12/blob/master/spm_list.m#L297).

It also uses `FWHM` 3 times (on lines [168](https://github.com/spm/spm12/blob/master/spm_VOI.m#L168), [177](https://github.com/spm/spm12/blob/master/spm_VOI.m#L177) and [199](https://github.com/spm/spm12/blob/master/spm_VOI.m#L199)) but I think these could be removed/worked around if needed.

To adapt the code from SPM as you suggested above these sections of the code would have to be modified/removed. The one thing I am confused about though as we don't currently have full brain clusterwise FDR. Also should the small volume corrections be available and/or parametric in the wild bootstrap display case?

TomMaullin:
This has now been fixed by PR #47

PR #47, opened 2018-09-07
Small volume corrections

TomMaullin:
This PR addresses issue #21 by adding the small volume option to the SwE display interface. It also addresses a (horrific) bug causing the display window to not reload correctly when a new option is selected.


swe_results_ui.m
```diff
@@ -600,10 +600,10 @@
         uicontrol(Finter,'Style','PushButton','String','small volume','FontSize',FS(10),...
             'ToolTipString',['Small Volume Correction - corrected p-values ',...
             'for a small search region'],...
-            'Callback','spm_VOI(SwE,xSwE,hReg);',...
+            'Callback','swe_VOI(SwE,xSwE,hReg);',...
             'Interruptible','on','Enable','on',...
             'Position',[015 095 100 020].*WS,...
-            'Enable','off')
+            'Enable','on')
```
nicholst: Minor point: `'Enable','on'` option is given twice.

TomMaullin:
Hi @nicholst ,

The bug is fixed!! And this PR is now ready for review! Please let me know if everything looks alright to you and, if possible, if everything runs correctly on your machine! Please let me know if you have any feedback/suggestions also!

nicholst:
Awesome! Was there a trick? A key “SPM GUI API lessons learned”?
-- 
Thomas Nichols, PhD

TomMaullin:
It turns out the problem was to do with the setup methods in `swe_results_ui`. The `"setup"` method of [`swe_results_ui`](https://github.com/NISOx-BDI/SwE-toolbox/blob/master/swe_results_ui.m#L234) is the first method to be called whenever a display is made. The error came from how the `"setup"` method does the following two steps:

 - Step 1: It first sets up the base outline of the GUI. This is done by calling `swe_results_ui` with the `"SetupGUI"` input argument on [line 331](https://github.com/NISOx-BDI/SwE-toolbox/blob/master/swe_results_ui.m#L331).
  - Step 2: It then, on [line 513](https://github.com/NISOx-BDI/SwE-toolbox/blob/master/swe_results_ui.m#L513), makes everything currently displayed in the GUI perminent. It does this by giving all of the child nodes of the GUI window the label `"PermRes"` (short for perminent results). The SPM functions are coded such that if the GUI is reset then the nodes with the label "PermRes" will not change (so for example page formatting etc).

The error here was caused by the table of results was being created in the wrong place. The table was being created in the `"SetupGUI"` method on [line 561](https://github.com/NISOx-BDI/SwE-toolbox/blob/master/swe_results_ui.m#L561) which was called during step 1. It was then being made permanent by step 2 and not being deleted when the display was reset. To overcome this I moved the call to `swe_list` to after step 2 (on [line 517](https://github.com/tommaullin/SwE-toolbox/blob/svc/swe_results_ui.m#L517)).

I guess the SPM GUI API lesson learned here would be that anything made in the "SetupGUI" method is here to stay for good. If you need something to change it shouldn't be created in this method.

nicholst:
Awesome!  Thanks so much!

TomMaullin:
Great - is this okay to merge?

nicholst:
I played around with this and it's great.  Could we, however, offer a warning that *if* WB has been run, no FWE inferences are available.  Something like

> WARNING: No FWE (voxel or cluster) results available for small volumes. Create a new WB analysis with a restricted analysis mask to obtain FWE results on a small volume. 

each time 'small volume' is clicked should avoid confusion.

TomMaullin:
Ah yes apologies! The warning must have slipped my mind, I will do this now!

TomMaullin:
Hi @nicholst ,

This warning should now be added. Please let me know if you have any further feedback/are happy to merge!

nicholst:
Super!  THanks, merge when ready.

nicholst:
Looks great! Fire away.

Issue #15, opened 2018-05-21
Add TFCE as inference option


nicholst:
Request from Mary Heitzeg's group at the University of Michigan

TomMaullin:
Hi @nicholst ,

I have been looking into how this could be done and I was wondering if the following is what you had in mind;

Built into the wild bootstrap option for SwE would be a `TFCE` option. This would be calculated by;

 - First, computing the `SwE{T/F}` map parametricly.
 - Second, converting the `SwE{T/F}` map to a TFCE statistic using a function like [this](https://github.com/markallenthornton/MatlabTFCE/blob/master/matlab_tfce_transform.m) (and saving the resultant map as `swe_vox_TFCE_c{c#}.nii`). Let this original TFCE statistic map be denoted `O` and `O_v` denote it's value at voxel `v`.
 - Third, performing a bootstrap which for each bootstrap, bootstrap number $I$, does the following:
   - Calculate the `SwE{T/F}` as usual. 
   - Convert the `SwE{T/F}` into a TFCE image (denote the TFCE image from the Ith bootstrap as `B_I`).
   - For every voxel, record:
![image](https://user-images.githubusercontent.com/22241320/45231278-7076c980-b2c3-11e8-926e-cf67269445ed.png)
      where `1()` represents an indicator function.



    - Record the global maxima in the TFCE image:
![image](https://user-images.githubusercontent.com/22241320/45231372-b03db100-b2c3-11e8-8093-26eedf7bb673.png)

 - Fourth, obtaining:
   - The uncorrected TFCE P values from:
![image](https://user-images.githubusercontent.com/22241320/45231567-4245b980-b2c4-11e8-9c18-4cba5c7d600a.png)
   - The FWE voxelwise corrected TFCE P values from looking at the distribution of the recorded maxima.
   - The FDR voxelwise corrected TFCE P values from performing: `spm_P_FDR` on the uncorrected TFCE P values.

The only additional choices the user should have is the variable `dh` for computing the TFCE integral.

The resultant output of this process should then be the maps:
 - `swe_vox_TFCE_c{c#}.nii` - The parametric TFCE image.
 - `swe_vox_TFCE_lp-WB_c{c#}.nii` - The TFCE uncorrected p value image.
 - `swe_vox_TFCE_lpFWE-WB_c{c#}.nii` - The TFCE voxelwise FWE corrected p value image.
[...]

PR #48, opened 2018-09-12
TFCE

TomMaullin:
This PR implements TFCE as discussed in issue #15.


TomMaullin:
Hi @nicholst ,

This PR is now review ready. Updates in this PR include:

 - TFCE computation (only for `.nii`/`.img` input - `.mat` I will leave for a different PR)
 - TFCE added to the GUI
 - TFCE results
 - An updated Travis test suite of 14 tests (4 parametric, 4 voxelwise wild bootstrap, 4 clusterwise wild bootstrap and 2 TFCE wild bootstrap - see the travis log for [more details](https://travis-ci.com/NISOx-BDI/SwE-toolbox/builds/85582379?utm_source=github_status&utm_medium=notification).

Please let me know what you think of the changes made! This is probably the largest edit I have made to the SwE code so, if you have the time at any point, please try out some of the options and use cases! I have gone through all the main `img`-input use cases twice (e.g. parametric voxelwise, WB voxelwise, WB clusterwise, WB TFCE, all for T and F) to check the displays all work correctly and tried to be as thorough as possible with selecting all possible options at the results stage but a second opinion would be much appreciated nonetheless!

nicholst:
On this TFCE PR, did you verify that you get precisely identical maps when `bwfuncexists` is true vs. false?

TomMaullin:
I did yes, I also ran it over and over quite a lot of times to triple check the maps were identical.

nicholst:
Great.  I still need to give it a spin with data..

TomMaullin:
Hi @nicholst ,

Apologies, yes you are completely right sorry - I made a mistake here and was reporting the FWE/FDR p value instead of the uncorrected equivalent... This should now be fixed - Please let me know if this all now seems correct to you. (Note: Just in case of anymore feedback, I haven't put this in our rc branch yet so you will have to checkout my TFCE branch to see the changes).

nicholst:
Thank you @TomMaullin !  This is great.

I just tested it.  For TFCE it looks fine.  For voxel FWE I get this:
```
Threshold: Height X = 48.55, p = 0.000; Extent k = 0 voxels.
vox F(5% FWE): 189.882, vox P(5% FDR): 0.028
```
and for (uncorrected) clusterwise (e.g. cluster forming p=0.0001 and k=10)
```
Threshold: Height X = 15.14, p = 0.000; Extent k = 10 voxels.
vox F(5% FWE): 189.882, vox P(5% FDR): 0.028
```

Can you remind me how the X and F thresholds relate here?

TomMaullin:
Hi @nicholst ,

The `Threshold: Height X=XXX` is the X statistic which is equivalent to the P value which was used for thresholding. As I can't give an exact value for this for bootstrapped FWE I have set this to be equal to the largest value in the X statistic map which was removed when FWE thresholding was applied. This is done [here](https://github.com/TomMaullin/SwE-toolbox/blob/TFCE/swe_getSPM.m#L870) . 

The `vox F(5% FWE): XXX` is the F value taken from the 95% percentile of the bootstrapped max(F) values (see [here](https://github.com/NISOx-BDI/SwE-toolbox/blob/master/swe_getSPM.m#L1180)).

nicholst:
In `swe_cp_WB.m` line 53, please update the help message to reflect the new TFCE images written out.

swe_cp_WB.m
```diff
@@ -619,6 +640,21 @@ function swe_cp_WB(SwE)
 
   end
   
+  %-Initialise parametric TFCE results images, if TFCE has been selected.
+  %---------------------------------------------------------------------- 
+  if TFCE
+      VlP_tfce_pos = swe_create_vol(sprintf('swe_tfce_lp-WB_c%02d%s', 1, file_ext), DIM, M,...
+                  'Original parametric TFCE statistic data.');
```
nicholst: The description isn't correct... this is "Uncorrected nonparametric TFCE p-value" or whatever matches a similar description for other similar files, right?

swe_cp_WB.m
```diff
@@ -619,6 +640,21 @@ function swe_cp_WB(SwE)
 
   end
   
+  %-Initialise parametric TFCE results images, if TFCE has been selected.
+  %---------------------------------------------------------------------- 
+  if TFCE
+      VlP_tfce_pos = swe_create_vol(sprintf('swe_tfce_lp-WB_c%02d%s', 1, file_ext), DIM, M,...
+                  'Original parametric TFCE statistic data.'); 
+      VlP_tfce_FWE_pos = swe_create_vol(sprintf('swe_tfce_lpFWE-WB_c%02d%s', 1, file_ext), DIM, M,...
+                  'Original parametric TFCE statistic data.');
```
nicholst: The description isn't correct... this is "FWE-corrected nonparametric TFCE p-value" or whatever matches a similar description for other similar files, right?

swe_cp.m
```diff
@@ -17,7 +17,8 @@ function swe_cp(SwE)
 % the following analagous maps:
 %
 %   - swe_vox_mask: The mask image for the analysis.
-%   - swe_vox_beta_c: The beta map for each contrast
+%   - swe_vox_beta_b: The beta map.
+%   - swe_vox_beta_c: The beta map for each contras.
```
nicholst: Let's call these 'contrasts' or 'con'... i.e. `swe_vox_con_c{c#}` (they're not beta's... they're linear combinations of betas)

nicholst: Honestly... I cannot see how `ess` exists for the F stat code (just poured over `swe_contrasts.m`... don't worry about it for F stats... just `con` for T stats.

TomMaullin: I have renamed the `beta_c{c#}` files but am unsure what you mean for me to do with the F statistic maps... what do you mean by `ess`?

nicholst: Don't worry about it.  For reference:  A F-test can be defined by
> F = (SSreg-SSreg0)/(SStot-SSreg) / ( (P-P0)/(N-P) ) = ESS/Err / (df1/df2)
i.e. the "extra sum of squares" explained by the full model relative to a reduced(null) model.  SPM writes out this "extra sum of squares", i.e. (SSreg-SSreg0) = ESS by analogy to the `con` images being the numerator of a T statistic.

I poured over the SWE code but I cannot find for my life an exact form for ESS... so we'll leave it for Bryan :)

TomMaullin: Ah right yes! I see now! Yes... I'm not sure on the exact form there either.

nicholst:
OK!  I finally checked out the footer... it looks good.  I'm happy with this beast of a PR!  Merge when you're ready and proceed with the 2.0.0 release ...

TomMaullin:
Okay brilliant! I'll go ahead with this!

PR #50, opened 2018-09-14
Bug for group size fixed

TomMaullin:
This PR aims to address a bug reported on the SwE mailing list. The bug occurred on [line 283](https://github.com/NISOx-BDI/SwE-toolbox/blob/master/swe_cp.m#L283) of `swe_cp`.

The reason for this bug was the variable `nSubj_g`. This is the variable which records the number of subjects in each group. E.g. `nSubj_g=[10, 12, 32]` would mean group one had 10 subjects, group 2 had 12 subjects and group 3 had 32 subjects.

The problem here was that the for loop on line 283 was meant to cycle through each subject in each group. I.e. for each group, with group number `g`, it was meant to cycle through `nSubj_g(g)` subjects (carrying on from the example above, for group 1 it should have cycled through 10 subjects, for group 2 it should have cycled through 12 subjects and for group 3 it should have cycled through 32 subjects).

The issue was that the `(g)` subscript was missing and as a result the loop was always trying to cycle through `nSubj_g(1)` subjects. This had not previously been caught as all of the examples in the testing suite have the same number of subjects in each group.

@nicholst , This PR is ready for review, please let me know if you have any feedback/everything here seems correct to you.

### DIFF
--- a/swe_cp.m
+++ b/swe_cp.m
@@ -280,7 +280,7 @@ if isfield(SwE.type,'modified')
         nSubj_g(g) = length(uSubj_g{g});
         uSubj_g_tmp = uSubj_g{g};
         
-        for k = 1:nSubj_g 
+        for k = 1:nSubj_g(g) 
 
              % The number of visits for subject uSubj_g(k)
              vis_g_subj(k) = sum(iSubj_g==uSubj_g_tmp(k));


nicholst:
Ooops!  That's a bug.  Thanks for quickly finding it. 

The only question I have: Do we know this was introduced?  It'd be nice to know if this bug could have let an analysis run to completion without errors, which is then very important for users to be notified that they need to re-run, and then let them know exactly whether they were affected by the problem.

TomMaullin:
I believe this was introduced when we introduced the footer information. The loop causing this error was recording the minimum and maximum visits for each group and therefore I think this was the only statistic that would have been affected.

My initial instinct would be that it isn't worth notifying users about as it is only the [min max] values in the footer for number of visits that would have been affected, and even then it is likely that if most subjects had roughly the same number of visits, then these values were probably at most underestimated by one or two. 

If you are happy with this change, shall I go ahead and merge and let the user know the code should now work for them?

nicholst:
Oh!  If it’s just for the footers, yes, that’s fine.  We just can note the nature of the bug when we make a release.

PR #58, opened 2018-10-02
Improve comment, so that FORMAT always lists outputs

nicholst:
Outputs were missing on many FORMAT lines.


nicholst:
Hi @TomMaullin,

This is a trivial fix on two files that had their outputs missing from the FORMAT line... a minor thing but a Matlab convention.  Please add this to the release.

-Tom

PR #60, opened 2018-10-17
[WIP] Implement SnPM-like control of random seed, adding swe_defaults

nicholst:


nicholst:
@TomMaullin: I need your help to modify the tests, so that, just like https://github.com/SnPM-toolbox/SnPM-devel/blob/master/test/test_snpm_pp.m they will disable seeding of the random number generator. 

Relevant code would look like:
```matlab
            % Random number generator should not be initialised with a
            % shuffled seed
            global SwEdefs
            SwEdefs.shuffle_seed = false;

```

TomMaullin:
Hi @nicholst ,

Okay sure, there are a few things I should probably note:

 - The tests are run in Octave not Matlab, and because of this the normal Matlab random number generator functions often don't work. It may be the case, as frustratingly I have found it often is with Octave random number generators, that the tests might just ignore your command resetting the seed anyway.
 - The tests are currently failing as they appear to be trying to call to an `snpm_get_defaults` function and not the random numbers - see the test log [here](https://travis-ci.com/NISOx-BDI/SwE-toolbox/jobs/152296866#L590). Until this is fixed it is hard for me to get an idea of exactly how Octave is handling the random number generator reset. I would fix it myself but I do not have permissions to edit your branch.
 - This PR is to the branch `new_release_2.0.0.rc`. I would suggest not doing this as this branch currently contains everything up to and including the last commit used in our beta release of the TFCE version of SwE. Keeping this branch in the state it's in currently will be extremely useful should anyone find any bugs/issues in the code as it means we can simply look at the code on this branch for reference. My suggestion would be that we recreate this PR but either going to `master` or a new branch on this repository named something appropriate, say `seed_reset` or something similar. What are your thoughts on this?

PR #62, opened 2018-10-17
Seed reset

nicholst:
SwE previously did not seed the random number generator, and this is a problem for Matlab in that identical WB results would be obtained from fresh Matlab sessions. (In Octave the RNG is randomly seeded each session).

This PR resolves this, randomly seeding the RNG by default, but allowing a disabling of the seeding via the `SwEdefs.shuffle_seed` global (important for testing).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ test/data/*
 test/MOxUnit/*
 # Autosaves
 *.asv
+.*~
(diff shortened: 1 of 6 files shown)


TomMaullin:
These changes all seem fine/correct to me and all tests pass! I'm happy to merge if you are!

PR #65, opened 2018-10-18
Updated data repository location

TomMaullin:
This is a quick PR which updates the location of the github data repository since we transferred the repository from `TomMaullin` to `NISOx-BDI`.


nicholst:
Merge when ready

PR #66, opened 2018-10-20
Incorrect quotes for MATLAB

justbennet:
This line was causing an error to be thrown when our researcher tried to run swe in MATLAB.


nicholst:
@TomMaullin should we add a hotfix to `new_release_2.0.0.rc` for this too?

PR #72, opened 2018-10-26
Correct some typos in help messages in design configuration

nicholst:


nicholst:
These are harmless changes... merging.

PR #75, opened 2018-10-29
Updates to 2.0.0rc

nicholst:


nicholst:
This is a mistake... it accidently includes countless updates that exist on master but not the new release.

Ug.  Deleting it.

Issue #74, opened 2018-10-28
SPM mode not set?


justbennet:
After completing the fMRI tutorial example, I chose the Save Batch and Script from the File menu, and I got a popup window that looked like

![save-error](https://user-images.githubusercontent.com/1595586/47617249-a05c7500-da9b-11e8-8180-13e9b81c84cb.png)

and in the MATLAB output window this message.

```Running 'Specify Model'
Mapping files                           :                        ...done

ans =

     1

Design configuration                    :                        ...done
Saving SwE configuration                :               ...SwE.mat saved
Design reporting                        :                        ...done
Done
Done    'Specify Model'
Done

Modality is not set: use spm('defaults','MOD'); where MOD is one of PET, FMRI, EEG.
```

Is that something that should be in the instructions or set by the SwE toolbox?

nicholst:
Huh... OK.  Can you tell me which version this is with? I.e. what you get with `swe('ver')`?

justbennet:
Tom,

This,

    bennet@mistakenhat:/usr/local/matlab/SwE-toolbox$ git show | head -1
    commit 622d715efd1618e3da76db8233616c24a02b2c31

and I believe I copied the modified version of the swe_ui_main.m from here

    https://raw.githubusercontent.com/nicholst/SwE-toolbox/FontTweak/swe_ui_main.m

If you think that the version of swe_ui_main matters, I can try to
replicate and track all the steps more closely.

-- bennet

nicholst:
PR #78 should address this.

PR #78, opened 2018-10-29
Set modality consistently

nicholst:
Modality is now set to FMRI (if not already set), in swe.m.  This is important if swe is directly called, without SPM being opened first.

Done to address issue #74.


TomMaullin:
This looks good to me - I am happy to merge if you are!

PR #82, opened 2018-11-02
Add cluster forming threshold to clustere header

nicholst:
The cluster forming threshold is an important parameter that could be lost... in lieu of awkwardly encoding it in the filename, put it in NIFTI description.


TomMaullin:
Hi @nicholst ,

This should be fine - the tests only check the actual values stored inside the NIFTI and not the header. Is there an existing convention for recording it in the description or should something like the below be fine:

'Non-parametric clusterwise FWE -log10(P) value data obtained by thresholding at p<{{p}}.'

TomMaullin:
This looks good to me! The tests should pass as only the values inside the NIFTI's are checked! I'm happy to merge if you are!

Issue #41, opened 2018-08-17
Add CIFTI support


nicholst:
Low priority, but growing due to growing use of HCP-style analyses.

See PALM for an template:
https://github.com/andersonwinkler/PALM/blob/master/palm_ciftiread.m
https://github.com/andersonwinkler/PALM/blob/master/palm_ciftiwrite.m

As a warning, it ain't pretty: Functionality is supported (as recommended by HCP folks) by system call to wb_command.

TomMaullin:
Hi @nicholst ,

Do you have any example data that could be used with these functions, e.g. some inputs to `palm_ciftiwrite`. My understanding is that, currently, the `.mat` input we have stores multiple volumes in some sort of triangular mesh. A secondary file is needed for the wild boostrap which stores structural information about the mesh. Would this be the same layout for `cifti` images, or could it be like `niftis` where there is one file per volume and the header includes the mesh information?

nicholst:
Can you connect with Habib and/or look at how he supports CIFTI in NINGA?  CIFTI, is a very general and complex format, and it is not feasible to support every facet of it, but he must have chosen some important subset of CIFTI types to support.

nicholst:
I have committed to provide cifti support by end of January.   @TomMaullin please ping me about this in Jan so we can talk with Habib and possibly Anderson Winkler about what we need to integrate PALM CIFTI code.

BryanGuillaume:
@nicholst, I am trying to work on this as you mentionned that it was relatively urgent. However, as I am not familiar with this type of data, I am not sure how people are using it. From what I can understand, it contains both volumetric data and surface data. Does the analysis is intended to be performed on both type of data in the same time or only one of them?

nicholst:
Hi @BryanGuillaume - Good points.  CIFTI is, in it’s whole, highly complex and can represent many many different types of data... surfaces, subcortical volumes and ‘connectomes’, square matrices where each row/column is a ROI. HOWEVER, in the first instance I was only thinking about surface data, and secondarily ‘brainordinates’ data, surface+subcortical.

Here are some links:
* https://mandymejia.com/2015/08/10/a-laymans-guide-to-working-with-cifti-files/
* http://mvpa.blogspot.com/2014/03/nifti-cifti-gifti-in-hcp-and-workbench.html

And, could I ask @andersonwinkler: What exactly do you support in PALM?  Just surfaces? Brainordinates? Connectomes too?

[...]

PR #96, opened 2018-12-22
[WIP] CIFTI Read/Write

TomMaullin:
This PR is a work in progress. It aims to address issue #41 by adding CIFTI support.


BryanGuillaume:
I believe this PR is not needed as we have added CIfTI support in PR #141. Thus, I am closing it without merging.

PR #98, opened 2019-01-14
Missing `dof_type`

TomMaullin:
This PR resolves an issue reported on the SwE mailing list in which a the case `dof_type=0` appears to not be correctly handled.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,13 +19,13 @@ install:
  - git lfs install  
  # Download data.
  - cd SwE-toolbox/test/data
- - if ! [ -d .git ]; then git clone https://github.com/NISOx-BDI/SwE-toolbox-testdata.git .; fi
+ - if ! [ -d .git ]; then git clone https://github.com/TomMaullin/SwE-toolbox-testdata.git .; fi
  # Delete any previous changes (retry because lfs might download files) 
- - git checkout master
- - travis_retry git reset --hard origin/master
+ - git checkout UOflowUpdate
+ - travis_retry git reset --hard origin/UOflowUpdate
  # A second time to allow for 3 more retries as "--retry 9" does not seem to be
  # taken into account by Travis CI
- - travis_retry git reset --hard origin/master
+ - travis_retry git reset --hard origin/UOflowUpdate
  - cd ../..
 jobs:
   include:
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ install:
  - cd SwE-toolbox/test/data
  - if ! [ -d .git ]; then git clone https://github.com/TomMaullin/SwE-toolbox-testdata.git .; fi
  # Delete any previous changes (retry because lfs might download files) 
- - git checkout -b UOflowUpdate
+ - git checkout UOflowUpdate
  - travis_retry git reset --hard origin/UOflowUpdate
  # A second time to allow for 3 more retries as "--retry 9" does not seem to be
  # taken into account by Travis CI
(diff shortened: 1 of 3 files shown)


TomMaullin:
Hi @nicholst ,

This PR is to fix the issue discussed on the SwE mailing list today... I believe the issue should be fixed, however on our toy example with 999 bootstraps, `Classic` SwE and `niave` degrees of freedom most of the brain appears to be p-values of zero. Does this seem correct to you?

image

nicholst:
No! Something is off. What are the EDF?

nicholst:
How can Z be inf? Didn’t we come up with a fix for that?

TomMaullin:
Hi @nicholst ,

The error df is "Niave" and equal to 22 but the contrast DF is only 1. As far as I can tell the infinite Z values/0 p values are coming from the original parametric analysis (as both the P and Z columns here are parametric). We did put a fix in place for zero p-values for the bootstrap p values I believe but not for the parametric.

To track down where exactly these infinite values occur I have found the lines on which P-values are calculated in my previous example analysis:

1. For each plane of the NIFTI image, the parametric T statistic values are calculated on [line 924](https://github.com/TomMaullin/SwE-toolbox/blob/210BugFix/swe_cp_WB.m#L924) of `swe_cp_WB.m`. The T statistic values seem reasonable for most voxels but some appear quite large (e.g. `T = 23.6958, 20.1759, 15.6964`). 
 2. The parametric T statistic is then given to the hypothesis testing function on [line 931](https://github.com/TomMaullin/SwE-toolbox/blob/210BugFix/swe_cp_WB.m#L931). Here each T value is converted to a P value on [line 2392](https://github.com/TomMaullin/SwE-toolbox/blob/210BugFix/swe_cp_WB.m#L2392). The large T statistic values are now converted into a P value of exactly 1.
 3. The p values for the current plane are converted to -log10(1-p) for the P value map on [line 963](https://github.com/TomMaullin/SwE-toolbox/blob/210BugFix/swe_cp_WB.m#L963) and Z scores for the Z map on [line 967](https://github.com/TomMaullin/SwE-toolbox/blob/210BugFix/swe_cp_WB.m#L967). The P values that were 1 now give a Z value of Inf. Similarly the p-values in the table (which are 1-p here) become 0.

This is where the Inf values in the table appear to be coming from. Does this seem as though it is correct behavior to you?

All of the Travis tests pass which means that for the 14 use cases we usually always check, this code has output exactly the same maps as the previous versions of the code. I am unsure if this "Inf" behavior is truly incorrect or whether it is just an artifact of converting T statistics to Z statistics.

nicholst:
Infinite Z’s can be unavoidable in SPM but you have to make sure your doing it in a way that minimizes underflow:

df = 20
T = 28
if T > 0
Z = -spm_invNcdf(spm_Tcdf(-T,df))
else
Z = spm_invNcdf(spm_Tcdf(T,df))
end

Manually, at the Matlab command line, see if the T values involved come out as Z=Inf. If yes, there’s nothing to do (except maybe implement Mark Woolrich’s t2z: http://www.fmrib.ox.ac.uk/datasets/techrep/tr00mj1/tr00mj1.pdf)

If no, something needs to be fixed. 

What I’m wondering is why we only found this now.

TomMaullin:
Hi @nicholst , The tests now pass and the number of Inf's in the output have been reduced following your previous suggestion. The updated screenshot for the analysis I showed previously is below:

image

Am I okay to merge this PR? (And the corresponding ground truth update on SwE-Toolbox-Testdata)

nicholst:
Yup!  Good to merge!

PR #110, opened 2019-05-20
[WIP] Fix a bug in subfunction swe_hyptest (in swe_cp_WB.m)

BryanGuillaume:
To compute the parametric p-value of an F-score of rank > 1, we need to multiply it by a factor equal to (edf - rankContrast + 1) / edf. In the subfunction swe_hyptest (in swe_cp_WB.m), this is correctly done using

`scoreTmp = (edf-rankCon+1) ./ edf .* score;`

Unfortunaltely, in some rare scenarios (e.g., a very low edf estimate and a high rank of the contrast), this factor can be negative for some voxels. This would yield some negative "scaled" F-scores, which is not theoretically possible (an F-score should always be positive or equal to 0). In order to circumvent this, swe_hyptest set all negative scaled F-score to 0:

` scoreTmp(scoreTmp < 0 ) = 0;`

In older versions of swe_cp_WB, we would then set the p-values of scaled F-scores that are equal to 0 to 1 using 

`p(scoreTmp == 0) = 1;`

For some reason, in the current version of the toolbox, it was changed to 

`p(scoreTmp == 0) = 0`

That does not seem correct as small F-scores should have high p-values. Now, I am suggesting to change it to 

`p(scoreTmp <= 0) = 1 `

This change sets the p-values back to 1 and modifies the strict equality to a "lesser than or equal to" comparison. The latter is used as an additional protection against float comparison equality faillure as I am not sure how Matlab handles float comparison for the value 0.

### DIFF
--- a/swe_cp_WB.m
+++ b/swe_cp_WB.m
@@ -2436,7 +2436,7 @@ else
     p(scoreTmp>0) = betainc((edf-rankCon+1)./(edf-rankCon+1+rankCon*scoreTmp(scoreTmp>0)),(edf-rankCon+1)/2, rankCon/2);
   else
     p(scoreTmp>0) = betainc((edf(scoreTmp>0)-rankCon+1)./(edf(scoreTmp>0)-rankCon+1+rankCon*scoreTmp(scoreTmp>0)),(edf(scoreTmp>0)-rankCon+1)/2, rankCon/2);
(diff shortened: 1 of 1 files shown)


nicholst:
Oh my!  A p-value of 0!  That can't be good.  

I'm good with change... merge away...

Issue #116, opened 2019-05-30
cluster-wise WB for deactivation seems to systematically yield empty results 


BryanGuillaume:
While doing some testing when I was implementing the post-hoc masking for the WB, I have noticed that the cluster-wise WB for deactivation (i.e. negative effect) seems to systematically yield empty results. As proof that something is off, I have run twice the same WB analysis but have used the contrast [-1 0 1] for the first analysis and [1 0 -1] for the second analysis. Here below, the obtained -log10(WB clusterwise p-value) images:

### contrast [-1 0 1] (left: positive effect, right: negative effect)
<img width="1428" alt="-1 0 1" src="https://user-images.githubusercontent.com/4426961/58624944-acc6db80-82d1-11e9-854d-9a26f39004f4.png">

### contrast [1 0 -1] (left: positive effect, right: negative effect)
<img width="1429" alt="1 0 -1" src="https://user-images.githubusercontent.com/4426961/58624942-acc6db80-82d1-11e9-8c4f-e934ffd9feb2.png">

We can indeed see that the negative effect images are empty for the 2 analyses while they should show something similar than in the other analysis.

Also, looking at the saved maximum cluster sizes, we observe that the registered max bootstraped cluster sizes are always 0 for the negative effect, which is unlikely to be correct.

### max cluster sizes in SwE.mat for [-1 0 1]
maxClusterSize: [353 5 5 4 144 8 9 8 4 8 11 2 14 17 6 17 42 12 7 5]
maxClusterSizeNeg: [1717 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]

### max cluster sizes in SwE.mat for [1 0 -1]
maxClusterSize: [1717 4 13 29 12 6 3 33 6 11 10 22 0 2 2 77 2 41 2 48]
maxClusterSizeNeg: [353 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]

Also, this issue seems to be present in the test data because the negative effect image is also empty.

The issue is likely to be somewhere in swe_cp_wb. I will try to fix this today.

BryanGuillaume:
I found the bug. For the bootstraps and image inputs, the variable indicating the voxels that were surviving the "negative" thresholding was initialised, but never updated:

### Initialisation of activatedVoxelsNeg
```
  if (SwE.WB.clusterWise == 1)
    activatedVoxels = false(1,S);
[...]

PR #117, opened 2019-05-30
Fix a bug in the cluster-wise WB for negative effects

BryanGuillaume:
This PR is a work in progress. It aims to correct issue #116, where the results of the cluster-wise WB for negative effects seemed systematicaly empty. The first reason for this is that the surviving voxels for negative effects had never been computed for each bootstrap (only for the original analysis), yielding maximum bootstraped cluster sizes of 0 for each bootstrap. This has now been fixed by recording correctly the surviving voxels at each bootstrap. The second reason why the results appeared empty was that the result image contained only nan or a single float value (the minimum WB p-value). SPM or FSLeyes does not seem to render these kind of images. This point has not been handled yet.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,13 +19,13 @@ install:
  - git lfs install  
  # Download data.
  - cd SwE-toolbox/test/data
- - if ! [ -d .git ]; then git clone https://github.com/NISOx-BDI/SwE-toolbox-testdata.git .; fi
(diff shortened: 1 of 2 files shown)


nicholst:
Looks good. FYI, FSLeyes (like FSLview before it) really struggles with NaN values, but that’s not our problem.

Issue #122, opened 2019-06-10
The display of clusterwise WB results does not seem correct


BryanGuillaume:
Working on the issue #103 (related to a post-hoc masking issue), I have noted that the display of results for clusterwise WB does not behave appropriately. 

Indeed, after running the results module form the batch system,

1. If I start by selecting "Activation", the result display seems correct. Then if I decide to switch to change the contrast to "Deactivation", I also get some display that looks ok. 
<img width="639" alt="Screenshot 2019-06-10 at 15 08 54" src="https://user-images.githubusercontent.com/4426961/59197885-b7ffee00-8b92-11e9-94d0-d805877e6290.png">
<img width="595" alt="Screenshot 2019-06-10 at 15 09 07" src="https://user-images.githubusercontent.com/4426961/59197884-b7ffee00-8b92-11e9-8c71-e49dcd72f39c.png">

2. If I start by selecting "Deactivation", I got an empty result. Then if I decide to switch to change the contrast to "Activation", I also get an empty result.

<img width="595" alt="Screenshot 2019-06-10 at 15 19 09" src="https://user-images.githubusercontent.com/4426961/59198058-22189300-8b93-11e9-9b3c-bfa9ebba4f19.png">
<img width="639" alt="Screenshot 2019-06-10 at 15 09 34" src="https://user-images.githubusercontent.com/4426961/59198059-22189300-8b93-11e9-9551-a53d292bc5c1.png">

These 2 set of result displays are contradictory. Also, we can note that the cluster threshold in case 1 is the same for deactivation and activation, which is suspicious. As the behaviour in case 2 is definitively wrong and the behaviour seems suspicious, I feel that there is a need to review all the code for the clusterwise WB display to ensure that it behaves appropriately.

BryanGuillaume:
I can confirm that the results for Deactivation in case 1 above are not correct as there should be 6 clusters with p=0.05 and not 3 like displayed. Nevertheless, the display result for Activation in case 1 seems correct. I fear that the cluster threshold used for Activation was also used for Deactivation

BryanGuillaume:
I can also see that the value clus k (5% FWE): 58 is the same for Activation and Deactivation (in case 1 and case 2). Using the infor in SwE.WB.clusterInfo, I can compute:
- For activation, P_FWE(clusterSize > 58) = 0.05
- For Deactivation, P_FWE(clusterSize > 24) = 0.05
This shows that there are also some errors for this part of the display.  

[...]

Issue #107, opened 2019-05-08
Use of "Change Significance" creates error WB results


nicholst:
At present, nothing stops a user from using the Change Significance menu... but of course this isn't possible for WB (or, at least, only the FWE threshold can be changed).

Trying to use it creates an error:

```
Index exceeds matrix dimensions.

Error in swe_getSPM (line 541)
      str  = xCon(Ic).name;

Error in swe_results_ui (line 245)
        [SwE,xSwE] = swe_getSPM(varargin{2});

Error in swe_results_ui>mychgcon (line 1232)
[hReg,xSwE,SwE] = swe_results_ui('setup',xSwE2);
 
Error while evaluating DestroyedObject Callback
```

BryanGuillaume:
I believe this has been fixed through the changes introduced in PR #125 and maybe also in PR #131. Currently, I cannot reproduced any error of this sort when I am using the "Change Significance" menu . Thus, I am closing this issue.

Issue #103, opened 2019-01-24
Post-hoc masking with nifti images does not appear to do anything


jacob-lee:
::SwE v.1.2.0
::SPM12

In Classic parametric SwE, explicit masking with a nifti image works if included in the SwE design, e.g.

`matlabbatch{1}.swe.smodel.masking.em = fullfile(masks_dir, 'gray_matter_mask.nii')`

However, post-hoc masking using the Results dialogue appears to do nothing.

Analysis
---------
The mask is never applied because the variable `Im` having the mask's path information is overwritten to `[]` before the masks are applied.

In swe_getSPM.m

`Mask = spm_input('apply masking','+1','b','none|contrast|image',[0,1,2],1);` 

solicits a mask (line 435). Choosing an image file sets `Im` to a cellstr with the image's path.

`[pf,nf,ef] = spm_fileparts(Im{1})` splits the mask's path to get mask name (line 527), but this appears to only be used to construct a label: `pf`, `nf` and `ef` are not otherwise used.

Then beginning on line 556, the variable holding the path to the mask `Im` is set to the empty matrix.

Therefore it is empty when, beginning on line 664, each mask applied in the for loop:

```matlab
for i = 1:numel(Im)
...
...
end
```

Commenting out lines 556-558 resolves my issue. I suspect doing so might cause an issue with wild bootstrapped analyses, given the preceding comment in the code.

BryanGuillaume:
@nicholst, I am currently working on this issue and I believe that I fix the bug reported. However, looking at the code, I am a bit suprise that the post-hoc masking is entirely disabled for WB analyses. Is there a particular reason for this? To my understanding, the post-hoc masking in SPM  (and in SwE) simply masks some voxels in the display but without modifying any scores or p-values. Thus, I guess that could also be used for the WB. Am I correct about this?

nicholst:
I can’t imagine why it is disabled. I don’t recall it as an action. @TomMaullin, do you have any memory of it?

TomMaullin:
[...]

PR #125, opened 2019-06-11
result display bug fixes and minor enhancements 

BryanGuillaume:
This PR mainly aims to correct a series of bugs in swe_getSPM that were affecting the display of results. They are listed below:

1. It corrects the bug reported in issue #103, where the pos-hoc masking for parametric test was not working anymore. This was due to the fact that a piece of code only dedicated to the WB was also systematically visited by parametric analyses. This was fixed by the use of an if-else condition  (see @@ -422,140 +434,131 @@ in swe_getSPM.m). 
2. It corrects a bug related to issue #122, where the deactivation results for clusterwise WB seemed erroneous. One of the reason for this was that the code was only considering the voxels surviving the positive contrast thresholding and hence, when starting a display of deactivation results, no voxel would survive (see case 2 in issue #122). The latter was fixed by appropriately selecting the voxel surviving the deactivation thresholding (see @@ -978,8 +979,13 @@).
3. It corrects another bug related to issue #122, where the 95th percentiles of maximum scores and maximum cluster sizes for negative t-contrasts was computed as those of the positive t-contrasts. See the modification at @@ -1258,11 +1287,23 @@ in  swe_getSPM.m
4. It corrects a bug in the UI of the results where clicking on some buttons in the **Contrasts** menu would crash the toolbox (see notably issue #107) or yield some weird result display as observed in issue #122. This behaviour was actually due to the fact that the variable SwE had never been saved after the contrast information had been appended to it. This has been fixed by saving it just after calling `SwE = swe_contrasts_WB(SwE);` (see @@ -288,7 +288,22 @@ and @@ -301,11 +316,8 @@). It is worth noting that this simple code modification has solved many missbehaviours that were observed, but it is dificult to see if it corrected them all appropriately. Some thorough checking might be needed.

This PR do also a bit of minor enhancements:

1. Modify some text printed in the matlab console (see @@ -1068,7 +1074,7 @@, @@ -1143,7 +1172,7 @@)
2. Fix some typo (see @@ -961,7 +962,7 @@)
3. Start implementing the post-hoc masking for WB. Nevertheless, the feature has been currently disabled and will be finished later in another PR when more urgent task will be done.

While this PR correct many bugs in swe_getSPM.m, I am pretty sure that there might be more issues (mainly enhancement) with the results display that should be fixed.

### DIFF
--- a/swe_getSPM.m
+++ b/swe_getSPM.m
@@ -288,7 +288,22 @@ end
 
 %-Get contrasts
 %--------------------------------------------------------------------------
-try xCon = SwE.xCon; catch, xCon = {}; end
+try
+  xCon = SwE.xCon; 
+catch
+  if isfield(SwE, 'WB') && ~exist('OCTAVE_VERSION','builtin')
+    SwE = swe_contrasts_WB(SwE);
+    % save SwE with xCon appended to it. This is important for future call of swe_getSPM for a specific Ic
+    if spm_check_version('matlab','7') >=0
+      save('SwE.mat', 'SwE', '-V6');
+    else
+      save('SwE.mat', 'SwE');
+    end
+    xCon = SwE.xCon;    
+  else
+    xCon = {};
+  end
+end
 
 try
     Ic        = xSwE.Ic;
@@ -301,11 +316,8 @@ catch
     % If we're in octave, assume we already have a contrast.
     elseif exist('OCTAVE_VERSION','builtin')
         Ic = 1;
-        xCon = SwE.xCon;
     % If we're doing WB, we already have a contrast. We just need to record it.
     else
-        SwE = swe_contrasts_WB(SwE);
-        xCon = SwE.xCon;
         if numel(xCon) == 2
             Ic = spm_input('Contrast Type','+1','b','Activation|Deactivation',[1,2],1);
         else
--- a/swe_getSPM.m
+++ b/swe_getSPM.m
@@ -1062,7 +1062,7 @@ if ~isMat
   end
   if isempty(Q)
       fprintf('\n');                                                      %-#
-      warning('SwE:NoVoxels','No voxels survive thresholding at p=%4.2f',pm);
+      warning('SwE:NoVoxels','No voxels survive thresholding');
   end
   
   % If we are doing clusterwise ask for threshold.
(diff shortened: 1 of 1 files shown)


BryanGuillaume:
@nicholst , Can I go ahead with this PR. While it may not fix everything for the WB result display, it does fix some important bugs. Also, I feel we should make a release after this PR because some  of the bugs that have been corrected in the few previous PRs seems quite important. For example, there was the bug for the computation of TFCE F-contrast and the bug for the deactivation of t-contast in the clusterwise WB.

nicholst:
Yes! And it will also give you experience with making a release. I’m pretty sure there are notes in the Wiki... let me know if not.

BryanGuillaume:
Indeed, there are some notes in the wiki. I will try to follow them. @nicholst, have you define some rules about the version number. The current one is 2.1.0 and the one before was 2.0.0. Thus, I am not sure if I should go for 2.2.0 or 2.1.1.

nicholst:
@TomMaullin - Do you remember the version numbering logic for SwE?  We went from 2.1.0 to 2.0.0, and Bryan is wondering what to use now?

TomMaullin:
Hi @nicholst,

I am a little confused... Is the current version not 2.1.0?

I tried to version the releases using a system like this:
https://en.m.wikipedia.org/wiki/Software_versioning

Roughly speaking I went up by 0.0.1 when I fixed bugs and made developer fixes and 0.1.0 when I added new functionality for the user into the code.

nicholst:
Sorry! I meant 2.0.0 -> 2.1.0. 

I like the distinction between bug fixes and new features. 

Bryan:  Is this enough to go on? And if so, can you edit the release instructions on the wiki to reflect this?

BryanGuillaume:
Thanks @TomMaullin 

I think I will go for a 0.0.1 increase for this one as there is no new feature added.

May I also ask you to clarify something about point 2 on the wiki:

> 2. Merge in all the unmerged updates that will be part of the release using git merge BRANCH_NAME where BRANCH_NAME is replaced by the name of the branch to be added to that release.

In my case, all the updates should be already merged into NISOx-BDI:master when I will create my release branch on my fork. Thus, do I need to merge anything as suggested by point 2? I guess this step is only used in case I would have branches on my fork that are not yet merged into NISOx-BDI:master and would need to be included into the release. Am I right?

@nicholst: I will also update the wiki reflecting the version numbering strategy.

TomMaullin:
This is actaully a wiki step that I took from other collaborations and was not my suggestion. I believe the reason behind this was so we could easily switch to a branch with the last release, if anything went wrong, without having to worry about looking at/pulling commit codes.

BryanGuillaume:
Thanks for the clarification. I am now merging this PR.

Issue #112, opened 2019-05-21
The bootstrapped TFCE scores for F-contrasts do not seem correctly computed


BryanGuillaume:
In `swe_cp_wb`, the original tfce score for an F-contrast is computed by first converting the original parametric -log10(p-values) into equivalent Z-scores and then use these equivalent Z-scores as input in `swe_tfce_transform`:

 ```
% Convert F statistics to Z scores.
      scorevol=-swe_invNcdf(10.^(-spm_read_vols(VlP)));
      scorevol(isnan(scorevol))=0;
          
      % Convert to TFCE.
      par_tfce = swe_tfce_transform(scorevol, H, E, C, dh);
```

This seems necessary as `swe_tfce_transform` appears to work with Z-scores as input as is indicated in the help of the function:

`% This function calculates the TFCE statistical from a Z statistic image.`

The **issue** seems to appear for the bootstrap tfce scores where, I believe, the input in `swe_tfce_transform` are not equivalent Z-scores but equivalent X-scores that are obtained through the function `swe_hyptest`. Here are the relevant piece of code:

### Computation of the bootstraped TFCE score
```
     % Calculate TFCE uncorrected p image.
      if TFCE
	% Obtain P values.
	hyptest=swe_hyptest(SwE, score, blksz, cCovBc, Cov_vis, dofMat);
	
	% Current XYZ indices
	currXYZ = XYZ(1:3, index);
	  
	  % T stat from this bootstrap
	  scorevol(sub2ind(DIM,currXYZ(1,:),currXYZ(2,:),currXYZ(3,:))) = hyptest.positive.conScore;
	
      end
      
    end % (bch)
    
    if TFCE

      if SwE.WB.stat == 'T'        
	
	% Bootstrapped tfce vol.
	tfce = swe_tfce_transform(scorevol,H,E,C,dh);
	tfce_neg = swe_tfce_transform(-scorevol,H,E,C,dh);    
        
      else
	
	% Bootstrapped tfce vol.
	tfce = swe_tfce_transform(scorevol,H,E,C,dh);
	
      end
```
### Computation of hyptest.positive.conScore in 
```
% Calculate converted score in a way which minimizes under/overflow
if (SwE.WB.stat == 'T')
    
    ....

else
    conScore = spm_invXcdf(negp, 1);
end
```

[...]

Issue #111, opened 2019-05-21
swe_update() cannot contact GitHub on my computer


BryanGuillaume:
swe_update() does not seem to work on my computer. Below the message I got on my Matlab:

```
>> swe_update()
Error using swe_update (line 24)
Can't contact GitHub
```
Note that, when typing the url that is in the code (https://github.com/NISOx-BDI/SwE-toolbox/releases/latest) in Safari or Google Chrome, it redirects me to the url https://github.com/NISOx-BDI/SwE-toolbox/releases/tag/v2.1.0.

I am not sure if the issue is only for some users like me (could be linked to a Matlab version or configuration) or if this is more global. @nicholst: Could you please test it on your computer to see if you can reproduce this issue?

Note that the same error appeared for SnPM:

```
>> snpm_update()
Error using snpm_update (line 21)
Can't contact GitHub
```

TomMaullin:
I can confirm that I do not get this error on my computer (just double checked) - I work on windows when I'm using Matlab... perhaps this is an OS thing?

BryanGuillaume:
@TomMaullin: Thank you for the feedback. This issue could indeed be an OS thing or due to a particular version of Matlab as they may have changed the problematic function `urlread` accross Matlab versions. What version of Matlab are you using? On my side, I am currently using Matlab R2017a on a Mac.

BryanGuillaume:
Reading some forums, it seems that this type of issue has occured for some other users of the function `urlread`. This [link](https://nl.mathworks.com/matlabcentral/answers/92506-how-can-i-configure-matlab-to-allow-access-to-self-signed-https-servers) seems to suggest that the issue may come from the fact that Matlab could not authentificate the HTTPS server certificate. Their proposed solution to overcome this issue seems quite cumbersome (need for the user to download the certificate, run a script,...). Thus, I would prefer to avoid this kind of solution. Nevertheless, I have tried to replace the function `urlread`by the function `webread` instead and it seems to  work for me... Thus, I will add a try catch statement in order to use `webread` instead of `urlread` if the latter fails to work.

TomMaullin:
I am running Matlab R2018b. `webread` sounds good. The only reason I used the page with the redirect is to avoid having to update it every time the code is updated.

BryanGuillaume:
This bug has been resolved in PR #115

Issue #99, opened 2019-01-17
SPM8 incompatible scan selection dialogue


jacob-lee:
Matlab 2015.a
SPM8
SwE (release 2.1.0)
-------------------
When selecting scans using the file dialogue in SwE (release 2.1.0), I get the following stack trace:

```
Operands to the || and && operators must be convertible to logical scalar values.
Error in cfg_files/subsasgn_check (line 41)
                    if strcmpi(item.filter,'image') || strcmpi(item.filter,'nifti')
Error in cfg_files/subsasgn (line 80)
            [ok val] = subsasgn_check(item,subs,val1);
Error in cfg_item/setval (line 55)
        item = subsasgn(item, substruct('.','val'), {val});
Error in cfg_util (line 876)
            cm = setval(cm, varargin{4}, false);
Error in cfg_ui>local_setvaledit (line 927)
    cfg_util('setval', udmodlist.cjob, udmodlist.id{cmod}, udmodule.id{citem}, val);
Error in cfg_ui>local_valedit_SelectFiles (line 956)
    local_setvaledit(hObject, val);
Error in cfg_ui>module_Callback (line 1416)
            local_valedit_SelectFiles(hObject);
Error in gui_mainfcn (line 95)
        feval(varargin{:});
Error in cfg_ui (line 53)
    gui_mainfcn(gui_State, varargin{:}); 
Error while evaluating UIControl Callback
```

With older versions of SwE, I got no such error.

Replacing spm8 with spm12 eliminates the error.

There is a difference in SPM's cfg_files/subasgn_check.m between spm8 and spm12 that accounts for the difference: namely that `strcmpi(item.filter,'image') || strcmpi(item.filter,'nifti')` is not executed is spm12. The proximate cause of the failure is that item.filter is not a scalar, e.g. {'image', 'mat'}, so that the result of the strcmp is a 1x2 vector.

nicholst:
@jacob-lee: Thanks so much for the detailed and insightful bug report.  While we do our best to maintain backwards compatibility with older versions of we don't have resources to do testing for anything but the current SPM12.

Anyway, I'm a bit mystified about the error... the code to select either 'image' or 'mat' has been around since last April -- see e.g. https://github.com/NISOx-BDI/SwE-toolbox/blob/v1.2.8/swe_cfg_design.m#L29 ... I'm not sure what happened with the latest change.

[...]

PR #113, opened 2019-05-22
Fix a a bug in the computation of TFCE scores for F-contrasts

BryanGuillaume:
This PR aims to correct the bug reported in issue #112. In brief, the bootstraped TFCE scores for F-contrasts were computed based on X-scores instead of Z-scores. This PR simply replace the X-scores by Z-scores.

PR #126, opened 2019-06-14
Release SwE Version 2.1.1

BryanGuillaume:
### Changes made since the previous version

- fix: Handle correctly the case `dof_type = 0` for Wild Bootstrap
- fix: Make the new version of the `SwE` scan selection dialogue compatible with `SPM8`
- fix: Fix a status message display for conjunctions
- fix: Fix a bug in subfunction swe_resid_corr for `dof_type = 1` (before, it would throw an error)
- fix: The subfunction swe_hyptest sets now the p-value of `F-Score <= 0` to 1 
- fix: The bootstrapped TFCE scores for F-contrasts are now computed based on Z-scores instead of X-scores 
- fix: Compute correctly the surviving voxels of each bootstrap of the cluster-wise WB for negative effects
- fix: Fix the post-hoc masking for parametric tests which was not masking any voxels anymore
- fix: Fix the results display of clusterwise Wild Bootstrap for negative contrasts which would sometimes use the voxels surviving the activation (i.e. positive contrast) 
- fix: Display correctly the 95th percentiles of maximum scores and maximum cluster sizes for negative t-contrasts (before, those of the positive t-contrasts were displayed)
- fix: Fix some errors in the `Contrasts` menu in the display window (before, clicking on some buttons in this menu would crash the toolbox)
- enh: save the `xCon` field in `SwE.mat` for Wild Bootstrap analyses
- enh: Prevent some Z-scores to be infinite
- enh: Improve the function `swe_update` by adding the possibility to use `webread` instead of `urlread` if the latter fails to fetch the most recent release version
- enh: Fix some typo 

### Link to relevant issues

- Issue #99
- Issue #103
- Issue #107
- Issue #111
- Issue #112
- Issue #116
- Issue #122

### DIFF
--- a/swe.m
+++ b/swe.m
@@ -11,7 +11,7 @@ function varargout = swe(varargin)
 % Written by Bryan Guillaume
 % Version Info:  $Format:%ci$ $Format:%h$
 
-versionNo = '2.1.0';
+versionNo = '2.1.1';
 
 try
   Modality = spm_get_defaults('modality');
--- a/swe_checkCompat.m
+++ b/swe_checkCompat.m
@@ -44,6 +44,7 @@ function swe_checkCompat(matVer, tbVer)
     % Record earliest compatible versions.
     earliestCompatVer('2.0.0') = '2.0.0';
     earliestCompatVer('2.1.0') = '2.0.0';
+    earliestCompatVer('2.1.1') = '2.0.0';
  
     % The below line works out the latest compatible version from the
     % earliest compatible versions. This code is now redundant but may be
--- a/swe_cp_WB.m
+++ b/swe_cp_WB.m
@@ -1740,7 +1740,7 @@ for b = 1:WB.nB
           % Get score volume from p values
           sv = -swe_invNcdf(hyptest.positive.p);
           % remove NaNs
-          sv(isnan(scorevol))=0;
+          sv(isnan(sv))=0;
           % Save as scorevol
           scorevol(sub2ind(DIM,currXYZ(1,:),currXYZ(2,:),currXYZ(3,:))) = sv;
         end


nicholst:
Oh! Wait! About 

> fix: The bootstrapped TFCE scores for F-contrasts are now computed based on X-scores instead of Z-scores

I think I may have misunderstood... the TFCE function expects Z values... what exactly goes in to the TFCE function?

BryanGuillaume:
Well spotted! That is a typo in this PR. The PR #113 actually does the opposite:

>This PR aims to correct the bug reported in issue #112. In brief, the bootstraped TFCE scores for F-contrasts were computed based on X-scores instead of Z-scores. This PR simply replace the X-scores by Z-scores.

To check this further, the new code in #113 is

```
     if strcmp(WB.stat, 'T')
           scorevol(sub2ind(DIM,currXYZ(1,:),currXYZ(2,:),currXYZ(3,:))) = hyptest.positive.conScore;
         % F test needs to be converted to Z
         else
           % Get score volume from p values
           sv = -swe_invNcdf(hyptest.positive.p);
           % remove NaNs
           sv(isnan(scorevol))=0;
           % Save as scorevol
           scorevol(sub2ind(DIM,currXYZ(1,:),currXYZ(2,:),currXYZ(3,:))) = sv;
     end
```
which should feed Z-scores in the variable scorevol. Now, looking at this code I have noticed an error in `sv(isnan(scorevol))=0;`. That should have been `sv(isnan(sv))=0;`. I will make a new commit to this PR correcting this.

I will also edit the text above.

nicholst:
Well spotted by you too! Thank you!

PR #129, opened 2019-06-26
Add the possibility to make voxel-wise WB inference based on param. Z- or X-scores

BryanGuillaume:
This PR aims to add the possibility to make voxel-wise WB inference based on parametric Z- or X-scores instead of using the raw F- or T-scores. The main reason to add this is to be somehow consistent with cluster-wise WB inferences because the formation of clusters is always done on these parametric Z- or X-scores.

Currently, the user can choose between both options. I can easily remove the "old way" (i.e. using raw F- or T-scores) and force the user to use Z- or X-scores now. @nicholst, do you have any preferences about this? 

An important note about this PR is that it also corrects a bug for the computation of bootstrapped minimum score. In the first versions of the toolbox, the code was

`minScore(b+1) = min(minScore(b+1), min(score));`

but was at some point changed into

`minScore(b+1) = min(score);`

This change is not appropriate because we are looping over chunks of voxels and we need to take the overall minimum and not the minimum of the last chunk.

Coming back to the option of using param. Z- or X-scores, I have run some tests to compare both options. Unfortunately, all my tests were actually yielding a constant edf over voxels, meaning that, in theory, the results should be the same. Here a pp-plot comparing the FWER p-values from both options in one of these tests:

Screenshot 2019-06-26 at 20 54 23

We can see that we can have some small discrepancies. I have checked and the reason is due to the tolerance of 10^-4 used when computing the WB p-values. As there is a shrinkage of the score towards 0 when converting to Z or X, this tolerance may have a different effect on the p-values.

I will try to generate some tests using a varying edf to see if there are more pronounced differences in these scenarios.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,13 +19,13 @@ install:
  - git lfs install  
  # Download data.
  - cd SwE-toolbox/test/data
- - if ! [ -d .git ]; then git clone https://github.com/NISOx-BDI/SwE-toolbox-testdata.git .; fi
+ - if ! [ -d .git ]; then git clone https://github.com/BryanGuillaume/SwE-toolbox-testdata.git .; fi
  # Delete any previous changes (retry because lfs might download files) 
- - git checkout master
- - travis_retry git reset --hard origin/master
+ - git checkout changeGT_test_wb
+ - travis_retry git reset --hard origin/changeGT_test_wb
  # A second time to allow for 3 more retries as "--retry 9" does not seem to be
  # taken into account by Travis CI
- - travis_retry git reset --hard origin/master
+ - travis_retry git reset --hard origin/changeGT_test_wb
  - cd ../..
 jobs:
   include:
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,13 +19,13 @@ install:
  - git lfs install  
  # Download data.
  - cd SwE-toolbox/test/data
- - if ! [ -d .git ]; then git clone https://github.com/BryanGuillaume/SwE-toolbox-testdata.git .; fi
+ - if ! [ -d .git ]; then git clone https://github.com/NISOx-BDI/SwE-toolbox-testdata.git .; fi
  # Delete any previous changes (retry because lfs might download files) 
- - git checkout changeGT_test_wb
- - travis_retry git reset --hard origin/changeGT_test_wb
+ - git checkout master
+ - travis_retry git reset --hard origin/master
  # A second time to allow for 3 more retries as "--retry 9" does not seem to be
  # taken into account by Travis CI
- - travis_retry git reset --hard origin/changeGT_test_wb
+ - travis_retry git reset --hard origin/master
  - cd ../..
 jobs:
   include:
--- a/swe_config_smodel.m
+++ b/swe_config_smodel.m
@@ -944,13 +944,29 @@ WB_infType.help    = {''
   ''
   'TFCE: Threshold Free Cluster Enhancement performed using the wild boostrap. (Note: this feature is not available for ''.mat'' input.'};
 
+
+% ---------------------------------------------------------------------
+% WB_isWbBasedOnParamXOrZ 
+% ---------------------------------------------------------------------
+WB_isWbBasedOnParamXOrZ         = cfg_menu;
+WB_isWbBasedOnParamXOrZ.tag     = 'WB_isWbBasedOnParamXOrZ';
+WB_isWbBasedOnParamXOrZ.name    = 'Parametric conversion of the statistics for voxel-wise inference';
+WB_isWbBasedOnParamXOrZ.labels  = {'no' 'yes'};
+WB_isWbBasedOnParamXOrZ.values  = { false true };
+WB_isWbBasedOnParamXOrZ.val     = { true };
+WB_isWbBasedOnParamXOrZ.help    = {  ' '
+                'no: the voxel-wise non-parametric p-values will be directly computed based on the T- or F-scores'
+                ' '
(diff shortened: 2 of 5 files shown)


nicholst:
Thanks for this! The equality of the two different approaches is
probably due to using a balanced ANOVA design. If you tried a
regression design (e.g. time only) or an imbalanced design, you'd
probably get different results.

I'm confused about your statement

> Currently, the user can choose between both options.

I don't recall this functionality, and, while we had only talked about
the option for FDR I didn't think you had implemented that.  Can you
clarify what you mean?  Is there a switch somewhere in the
configuration?  Or do you mean users see both in the results?

BryanGuillaume:
>  Currently, the user can choose between both options.

That only means that the choice between both approaches has been added in the "specify model" batch module in this PR (see screenshot below).  The main purpose of this is to be able to easily test and compare both approaches right now.  After more testing, we can decide to remove this choice option in the batch if we want to force the user to one particular approach.

I will test today with more complicated models to see how this affects the results.

Screenshot 2019-06-27 at 11 18 05

BryanGuillaume:
Here are some results from the same analysis but with some missing data (33%) and 99 bootstraps. We can now observe some differences between voxel-wise p-values. Here below some scatterplot of p-values and the histograms of the difference between uncorrected p-values and between FWER p-values. 

### Scatterplot of unc. p-values (94% of the voxels are in the middle line)
Screenshot 2019-06-27 at 15 40 05

### Histogram of unc. p-value difference (cropped in y for the central bin containing 94% of the voxels)
Screenshot 2019-06-27 at 15 19 13

### Scatterplot of FWER p-values (98% of the voxels are in the middle line)
Screenshot 2019-06-27 at 15 38 24

### Histogram of FWER p-value difference (cropped in y for the central bin containing 98% of the voxels)
Screenshot 2019-06-27 at 15 20 47

Thus, there is a match for the majority of the voxels, but some discrepancies for the remaining. It is hard to conclude whether one method is better than the other without doing some simulations.

BryanGuillaume:
@nicholst, can I go ahead and merge this PR. In brief, it forces the WB to use parametric Z or X-scores instead of the raw T or F-scores as bootstrapped scores. I have also updated the test data in accordance to this change.

nicholst:
Sorry for the delay @BryanGuillaume !

We agreed to *remove* the option for users to select this, right?  

Don't you think that for future maintenance of the code it's better to remove the (unused) code in `swe_config_smodel.m` and all the conditionals in the other functions?

BryanGuillaume:
Yes, we agreed to remove the option for users to select this.

For maintenance of the code, I am fine to remove all the extra code and I will do this now.

The only annoying thing that will come from this PR is for the display of results (that I am currently updating for another future PR) of older WB analyses. Indeed, for voxel-wise FWER WB inferences, it will be stated the Z or X height threshold (coming from the (1-alpha)th percentile of the max Z or X score). That will be correct for new analyses as the score are now Z or X, but not for older analyses where the max scores were T or F.  Thus, I am not sure if I should leave a tag somewhere in SwE.mat to detect whether the analysis was done prior this PR or not, in which case, I would simply use 'T' or 'F' instead of 'Z' or 'X' as the thresholded score in the display of results. Do you think that would be a good solution? Or do you have another idea?

nicholst:
Very good point!  We have committed to backwards compatibility of saved results until we have a release that changes the major version number (the first digit in x.x.x), so I would be grateful if you do indeed flag this, and display results from before this PR appropriately.  

Then, also make an issue: "Changes for Version 3.x", which reminds us to drop the compatibility code.  We can also use that issue as a random collection of things to do in future versions.

BryanGuillaume:
Ok, I have removed the extra code. Now, the two files `swe_config_smodel.m` and  `swe_run_smodel.m` are reverted back to their original state before this PR and there are only changes in swe_cp_wb.m. Basically, now,  all the T- and F-scores are simply substituted by Z and X-scores when we compute the WB p-values and max. (and min.) scores. I have also added a flag in SwE.WB that indicates the analysis was run after this PR. That will be useful to correctly display the results. 

As an additional reminder, this PR also corrects a bug in the computation of bootstrapped minimum scores (see explanation above in the first message) and also computes the uncorrected p-values using the "tolerance trick" (i.e. replace `a >= b` by `a > b - tol`)

As soon as Travis confirms the last commits pass the checks, I will merge this PR.

nicholst:
One last thought... do you need a special flag in SwE.WB?  Shouldn't we
record / have set somewhere the current version of SwE?  Like something in
SwEdefs?  Then you could simply check that the version is greater than X to
know what to do?  Seems better than a special flag, as we might need a flag
latter to do something else special.

BryanGuillaume:
Yes, I agree that using the model version would be a better alternative.

Looking at the code, the version is set after specifying the model. Thus it should be ok to use this info in the majority of the scenario. However, we could imagine a scenario where the user would have specified his model in an older version of the toolbox and run it with a new version of the toolbox. In this scenario, the toolbox would believe that the model was estimated through an old version of the toolbox. Note that, while that might appear to be a very unlikely scenario, that could easily happen for users that modify an old SwE.mat template before running it using swe_cp(SwE) (I am actually often doing this). 

One solution for this would be to reupdate the version after estimating the model or to throw an error message before running the model if `SwE.ver ~= swe('ver')`, forcing the user to use the same version of the toolbox for specifying and estimating the model. @nicholst , do you have any opinion about this or am I overthinking things?

nicholst:
I *think* this is simple: you set the version anytime you write out a SwE.mat.  If it’s not set we know the SwE.mat is from before this PR. Right?

BryanGuillaume:
Before this PR, it is already set, but after the specification module. Nevertheless, I can set it again after the estimation module. I do not think that would cause an issue.

nicholst:
Ok!
-- 
Thomas Nichols, PhD

PR #131, opened 2019-07-08
Modify the thresholding and the reporting of WB results

BryanGuillaume:
The main goal of this PR is to modify the way the thresholding and the display of results are done for the Wild Bootstrap. Some additional improvements, not necessarily related, have also been done. Below is a brief summary with explanations of the changes:

### Changes mainly related to the thresholding
- The WB thresholding is now inclusive when done on the p-values and exclusive when done on the score.
- The WB score thresholds are now computed as the (1-alpha)th percentile of the score distributions. Before it was computed as the largest score not surviving the thresholding.
- The thresholding for voxel-wise uncorrected inference is now done on the WB uncorrected p-values instead of the parametric uncorrected p-values.
- Some piece of code has also been modified in such a way that the WB thresholding can be done on images that would be masked by a post-hoc masking. In particular, we use now the variable `XYZ` instead of the variable `XYZum` Note that this post-hoc masking is currently disabled for WB, but might be enabled in a future PR. 
- swe_contrasts_WB now links to the WB uncorrected p-values instead of the parametric uncorrected p-values. This is necessary to threshold and display the WB uncorrected p-value instead of the parametric p-values.
- In swe_getSPM, we check if SwE.xCon links to the WB uncorrected p-values instead of the parametric p-values. If this is not the case, we correct this and overwrite SwE.mat. This is necessary for scenarios where SwE.xCon had been produced using an old version of swe_contrasts_WB.

### Changes mainly related to the display of results
- We now do not report a score threshold for voxel-wise WB inferences as these are voxel-dependent. Before, the parametric score threshold was shown.
- The score threshold for FWER voxel-wise inference is reported as a T/F-score for models estimated using a version older than or equal to 2.1.1. It is reported as a Z/X-score otherwise. This is to comply with PR #129.
- Now the WB p-value treshold are reported as inclusive and the WB score threshold as exclusive.
- The WB cluster-wise FWER p-values are now only reported if this a WB cluster-wise FWER is requested by the user.
- The left side of the footer has been reduced in width to avoid text collision with the right side when the threshold info of a WB cluster-wise inference is reported
- When a threshold is specificaly from the WB, it is indicated now that this a Wild Bootstrap threshold. For example, `Extent threshold` --> `Wild Bootstrap extent threshold`.

### Changes mainly related to the computation of statistics
- The TFCE  and cluster-wise WB FWER p-values are now computed using the tolerance trick (i.e. `a >= b` is replaced by `a > b - tolerance`).
- The maximum TFCE scores are now saved into `SwE.WB.TFCE.maxTFCEScore` and `SwE.WB.TFCE.maxTFCEScore_neg`(before they were not recorded).

### Miscellaneous 
- A new function `swe_compareVersions` has been added. It can compare version numbers of type x.x.x and is currently used to detect whether the voxel-wise WB p-values were based on T/F-scores or on F/X-scores.
- Typo fix: `Mn` is replaced by `Min`

Note that this PR is relatively hard to check as this is mainly related to the way the results are displayed. I have already checked this several times if all the changes looks ok to me. Tomorrow, I will check this again one more time and append some screenshots showing the changes for each type inference (before/ after the PR). That will help us to check if the changes are indeed useful and correct.

### DIFF
--- a/swe_checkCompat.m
+++ b/swe_checkCompat.m
@@ -86,7 +86,7 @@ function lcv = latComVer(ecv)
             if ~isKey(lcv, ecv(ver))
                 lcv(ecv(ver)) = ver;
             else
-                if vercheck(lcv(ecv(ver)), ver, '<=')
+                if swe_compareVersions(lcv(ecv(ver)), ver, '<=')
                     lcv(ecv(ver)) = ver;
                 end
             end
@@ -102,35 +102,4 @@ function lcv = latComVer(ecv)
         end
     end
     
-end
-
-% Perform <=, <, >, >= on version numbers.
-function vcheck = vercheck(ver1, ver2, flag)
-
-    % Split version number into parts.
-    ver1 = strsplit(ver1, '.');
-    ver2 = strsplit(ver2, '.');
-    
-    % Work out base for comparison.
-    base = max(cellfun(@(a) str2num(a), {ver1{:} ver2{:}})) + 1;
-    
-    % Convert to num.
-    ver1 = cellfun(@(a) str2num(a), {ver1{:}});
-    ver2 = cellfun(@(a) str2num(a), {ver2{:}});
-    
-    % Make into number in base `base`
-    ver1 = ver1(1)*base^2 + ver1(2)*base^1 + ver1(3)*base^0;
-    ver2 = ver2(1)*base^2 + ver2(2)*base^1 + ver2(3)*base^0;
-    
-    if strcmp(flag, '<=')
-        vcheck = (ver1 <= ver2);
-    elseif strcmp(flag, '<')
-        vcheck = (ver1 < ver2);
-    elseif strcmp(flag, '>')
-        vcheck = (ver1 > ver2);
-    elseif strcmp(flag, '>=')
-        vcheck = (ver1 >= ver2);
-    else
-        error('unknown flag')
-    end
 end
--- a/swe_compareVersions.m
+++ b/swe_compareVersions.m
@@ -1,0 +1,42 @@
+function result = swe_compareVersions(ver1, ver2, comparisonOperator)
+% Compare two version number of the type 'x.x.x'
+% =========================================================================
+% FORMAT: result = swe_compareVersions(ver1, ver2, comparisonOperator)
+% -------------------------------------------------------------------------
+% Inputs: 
+%   - ver1: first version number of the type 'x.x.x'
+%   - ver2: second version number of the type 'x.x.x'
+%   - comparisonOperator: comparison operator ('<', '<=', '==', '>=' or '>')
+% =========================================================================
+% Author: Bryan Guillaume & Tom Maullin (05/07/2019)
+% Version Info:  $Format:%ci$ $Format:%h$
+  
+  % Split version number into parts.
+  ver1 = strsplit(ver1, '.');
+  ver2 = strsplit(ver2, '.');
+  
+  % Work out base for comparison.
+  base = max(cellfun(@(a) str2num(a), {ver1{:} ver2{:}})) + 1;
+  
+  % Convert to num.
+  ver1 = cellfun(@(a) str2num(a), {ver1{:}});
+  ver2 = cellfun(@(a) str2num(a), {ver2{:}});
+  
+  % Make into number in base `base`
+  ver1 = ver1(1)*base^2 + ver1(2)*base^1 + ver1(3)*base^0;
+  ver2 = ver2(1)*base^2 + ver2(2)*base^1 + ver2(3)*base^0;
+  
+  if strcmp(comparisonOperator, '<=')
+    result = (ver1 <= ver2);
+  elseif strcmp(comparisonOperator, '<')
+    result = (ver1 < ver2);
+  elseif strcmp(comparisonOperator, '==')
+    result = (ver1 == ver2);
+  elseif strcmp(comparisonOperator, '>')
+    result = (ver1 > ver2);
+  elseif strcmp(comparisonOperator, '>=')
+    result = (ver1 >= ver2);
+  else
+    error('unknown comparison operator')
+  end
+end
--- a/swe_contrasts_WB.m
+++ b/swe_contrasts_WB.m
@@ -65,7 +65,7 @@ function [SwE] = swe_contrasts_WB(SwE)
     % Add the SwE volumes.
     %----------------------------------------------------------------------
     DxCon.Vspm = spm_vol(sprintf('swe_vox_%c%cstat_c%.2d%s', eSTAT, STAT, 1, file_ext));
-    DxCon.VspmUncP = spm_vol(sprintf('swe_vox_%cstat_lp_c%.2d%s', STAT, 1, file_ext));
+    DxCon.VspmUncP = spm_vol(sprintf('swe_vox_%cstat_lp%s_c%.2d%s', STAT, wbstring,1, file_ext));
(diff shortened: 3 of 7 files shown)


BryanGuillaume:
In order to visualise the effect of this PR, here below is an example of result desplay before and after this PR.
### Voxel-wise FWER WB inference (Before/After this PR)

- Before, the p-values equals to the threshold (here 0.1) would not survive.
After, the p-values equals to the threshold (here 0.1) would survive.
- Before, the height threshold would be computed as the highest Z-scores not surviving the thresholding.
After, the height threshold would be the (1-alpha)th percentile of the max distribution.
- Before, the thresholds would be denoted using `=` or `<`.
After, the thresholds would be denoted using `>=`, `<=`, `>` or `<`.
- Before, the cluster-wise FWER p-values corresponding to the peak voxels would be displayed.
After, the cluster-wise FWER p-values corresponding to the peak voxels would not be displayed.
- Before, the 5% FWER cluster size would be displayed in the footer.
After, it is not displayed (because it is only valid for a specific forming threshold, which is not unlikely to be the one displayed).
- Before, the uncorrected p-values in the table would be the parametric version.
After, the uncorrected p-values in the table would be the non-parametric WB version.
- Before, the footer would be composed of 2 columns of the same size.
After, the second column is smaller in order to be able to display wider text in the first row of the first column


### Before
vox_FWER_WB_before

### After
vox_FWER_WB_after

nicholst:
Super!  But, just to be clear, *if* you had used a cluster-forming
threshold that coresponded to one previously saved by WB, you *would* get
cluster-level p_FWE-corr... right?

BryanGuillaume:
No, because the thresholding would not be exactly the same. Indeed, the cluster forming thresholding during the computation of FWER cluster-wise p-values is always done on uncorrected parametric p-values. In the result section, the cluster-forming thresholding would be done on the uncorrected WB p-values, which may yield different results (i.e. different clusters). For example, while we could expect some agreement, we could imagine that a parametric cluster could be split in two after using WB p-values,... I feel that would be risky to then display the cluster-level FWER p-values. Another scenario would be the case of selecting the classic SwE with Approx III. Indeed, in this scenario, the parametric p-values tend to be overestimated (particularly in small samples). Thus, the clusters are likely to be smaller with parametric p-values than with WB p-values.

Currently, the cluster-level FWER p-values are only shown when the user specifically asks for a cluster-wise FWER thresholding.

BryanGuillaume:
### Cluster-wise FWER WB inference (Before/After this PR)
The differences are similar to the voxel-wise FWER WB inference scenario above. More specificaly to this type of inference,

- Before, the reported cluster threshold was actually inclusive (i.e. it was the size of the smallest cluster surviving the thresholding). Also note the discrepancies between the 2 first rows of the first column in the footer when controlling at 5% FWER (stating and inclusive 60 voxels vs and exclusive 32 voxels)
After, it is exclusive and taken as the (1-alpha)th percentile of the max distribution. Note that the two values in the footer now match as one would expect.
- Before, as can be observed in the scenario with 2% FWER, the p-value threshold was exclusive.
After, it is inclusive.
- Before, the cluster size p-value threshold was not in the footer
After, it has been added

### Before for FWER control at 5%
clus_FWER_WB_before

### After for FWER control at 5%
clus_FWER_WB_after

### Before for FWER control at 2%
clus_FWER_WB_before2

### After for FWER control at 2%
clus_FWER_WB_after2

BryanGuillaume:
### Voxel-wise FDR WB inference (Before/After this PR)
Similar comments than above +

- Before, the thresholding was done on the parametric FDR p-values while the FDR p-values displayed were from the WB (see the surviving voxel with WB FDR p = 0.137 while thresholding at 0.1). 
After, the thresholding is done on the WB FDR p-values and all the surviving voxing are well under or equal to the threshold.
- Before the parametric score threshold was reported.
After, no score threshold is reported because they are voxel dependent.

### Before
vox_FDR_WB_before

### After
vox_FDR_WB_after

BryanGuillaume:
### Voxel-wise Uncorrected WB inference (Before/After this PR)
Similar comments as above. More specifically,

- Before, the thresholding and displayed p-values were based on the parametric uncorrected p-values.
After, the thresholding and displayed p-values are based on the WB uncorrected p-values.
- Before, the parametric score threshold was shown.
After, no WB score threshold is shown beacause it varies form voxel to voxel.

### Before
vox_unc_WB_before

### After
vox_unc_WB_after

nicholst:
These comparisons are super helpful.  Thank you!

BryanGuillaume:
### Cluster-wise Uncorrected WB inference (Before/After this PR)
Similar comments as above. In particular,

- Before, the height thresholding was done based on parametric uncorrected p-values.
After, it is done based on WB uncorrected p-values. Note that in both cases (before and after), the cluster size threshold given by the user is inclusive.

### Before
clus_unc_WB_before

### After
clus_unc_WB_after

BryanGuillaume:
### TFCE WB inference (Before/After this PR)
Similar comments as above. In particular, 

- Before, the p-value TFCE thresholding was actually exclusive. Interestingly, in the example below, the value 0.05 seems to pass the thresholding while it should not. I believe that is due to the numerical conversion between p and -log10(p), and then back to p. Indeed, extracting a corresponding -log10(0.05) from the saved image (referred as "lp") and looking at the difference, we have 10^-(lp) - 0.05 = -5.0799e-09 (Also, on my computer, in Matlab,` 10^-(-log10(0.05)) - 0.05 = -1.3878e-17`)
After, the p-value TFCE threshold is inclusive

### Before
tfce_WB_before

### After
tfce_WB_after

BryanGuillaume:
I believe the 6 scenarios above are a good summary of all the cases we can encounter in practice. 

Nevertheless, I would like also to mention the following behaviour, which has not been changed but bothers me a little. The reported TFCE p-value in the table is the smallest in the corresponding reported cluster while the reported peak statistics on the same row are those of the voxel in the cluster with the highest Z-score. Thus the voxels are not necessary the same.

Similarly, for voxel-wise uncorrected and FDR WB inference, I believe the peak voxel is decided using the Z-score and may not correspond to the voxel with the smallest WB p-value in the cluster.

nicholst:
> Nevertheless, I would like also to mention the following behaviour, which has not been changed but bothers me a little. The reported TFCE p-value in the table is the smallest in the corresponding reported cluster while the reported peak statistics on the same row are those of the voxel in the cluster with the highest Z-score. Thus the voxels are not necessary the same.

There is an important property of TFCE:  Any isocontour in a TFCE image can also found in the original test statistic image.  This also means that for every local maxima in the TFCE image you will find a local maximum in the corresponding T/F image (up to 'vertical' resolution limits from using a finite number of thresholds).  Have you observed that the local maxima in the TFCE are different (substantially different) from the T statistic image?

But, anyway, I appreciate the note and agree that the disconnect between the right-hand columns and the TFCE inferences on the left are not ideal.

BryanGuillaume:
Oh yes, I remember now this TFCE property. The concern came from the code which extracts the TFCE and peak statistics separately, so could potentially extract the statistics from 2 different voxels. Nevertheless, this should not be an issue thanks to this TFCE property.

nicholst:
Just to be clear, due to the finite number of levels h, TFCE might have
some 'flat spots', where there are multiple voxels with equal TFCE scores,
but the local maxima in the T image *would* be the true TFCE peak if we
used more and more levels h.  So, anyway, it is faithful to report T/F
peaks as TFCE peaks even if there isn't a unique peak in the TFCE image.

nicholst:
Good to merge

Issue #132, opened 2019-07-11
The display of results for the classic parametric SwE crashes


BryanGuillaume:
The display of results for the parametric classic SwE crashes when I am trying to display some results, just after computing the SPM images and just before displaying the results. Here below the error message:

```
11-Jul-2019 15:20:48 - Running 'Results Display'
	contrast image  3               : ...written swe_vox_con_c03.nii
	spm{T} image  3                 : ...w......wri...written swe_vox_edf_c03.nii
	SPM computation                 :                        ...done
11-Jul-2019 15:21:01 - Failed  'Results Display'
Operands to the || and && operators must be convertible to logical scalar values.
In file "/Users/bryanguillaume/Documents/GitHub/SwE-toolbox/swe_list.m" (???), function "swe_list" at line 506.
In file "/Users/bryanguillaume/Documents/GitHub/SwE-toolbox/swe_list.m" (???), function "swe_list" at line 163.
In file "/Users/bryanguillaume/Documents/GitHub/SwE-toolbox/swe_results_ui.m" (???), function "swe_results_ui" at line 527.
In file "/Users/bryanguillaume/Documents/GitHub/SwE-toolbox/swe_run_results.m" (???), function "swe_run_results" at line 22.

The following modules did not run:
Failed: Results Display
```

More precisely the error is thrown at the line of code:

`if xSwE.dofType~=0 && diff > 10^-10`

because diff is actually empty. Tracing back through the code why it is empty, it seems that this comes from the fact the variable `Vedf` is not saved in the variable `xSwE`.

I will try to resolve this issue by saving this variable into xSwE and see if this works properly after that.

BryanGuillaume:
This issue has been fixed in PR #133.

PR #133, opened 2019-07-12
Fix the result display error observed for the classic parametric SwE

BryanGuillaume:
The source of the error (see issue #132 ) seems to be due to the fact that the appropriate information about the number of degrees of freedom was not saved correctly for the classic parametric SwE in swe_getSPM.

In brief, before this PR, the variable edf was saved into xSWE for all parametric analysis and Vedf was saved into xSWE only for all analysis using the modified SwE. These 2 conditions seem weird as the scalar value edf is only relevant for a naïve estimation of the numbers of degree of freedom that is assumed constant across the brain (i.e. when SwE.dof.dof_type == 0). In addition, in all the other cases (i.e. when SwE.dof.dof_type > 0), the numbers of degree of freedom is potentially varying accross voxels and saved into the image Vedf. Thus, in this PR, we use now a condition on SwE.dof.dof_type to decide which variable to save.

This PR seems to fix the issue #132.


nicholst:
Good to merge

PR #136, opened 2019-07-26
use the resampling matrix in SwE.WB if it exists

BryanGuillaume:
The goal of this PR is to add an "hidden" option to supply a WB resampling matrix instead of letting the toolbox to randomly generate one. This is achieved by adding to SwE.WB a field SwE.WB.resamplingMatrix with the WB resampling matrix. This is "hidden" in the sense that the user must added manually through code and not though the toolbox itself. I made this option "hidden" in order to limit the possibility for some users to feed a "wrong" WB resampling matrix. 

Note that this PR is very useful in order to test the toolbox and compare its results to those of the fsl version.

### DIFF
--- a/swe_cp_WB.m
+++ b/swe_cp_WB.m
@@ -1543,13 +1543,21 @@ end
 %- Produce bootstraps and maximum stats 
 %==========================================================================
 
-% Produce the random value following the Rademacher distribution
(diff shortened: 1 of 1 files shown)


BryanGuillaume:
I am merging this now because I need this to benchmark and update accordingly the FSL and SPM versions. We can always decide later to modify what this PR is doing if needed.

PR #137, opened 2019-07-29
Add the number of degrees of freedom Approximation I for the WB

BryanGuillaume:
The goal of this PR is to add the option to use `Approximation I` as a usable estimator for the number of degrees of freedom when a WB analysis is performed. Before this PR, this was not implemented and an error message was thrown to indicate this to the user. 

I have tested this PR for the classic SwE using the image format and the results match those of the fsl version of the toolbox. Thus, I can assume that this PR is correct for this particular scenario. Nevertheless, the code is also implemented for the modified version of the SwE and when the input is a ".mat" format. Those cases would need to be checked before merging this PR.

Note also that the code is not very pretty because I had to do it quickly in order to be able to check the fsl version of the toolbox on this type of estimator for the classic SwE. In particular, the edf is computed alongside the contrasted SwE computation because we need to have access to each group (for the modifed SwE) or subject (for the classic SwE) contribution. These contributions were only available during the contrasted SwE computation. Therefore, they are not available when the subfunction swe_hyptest is called. Thus, swe_hyptest cannot be used without a strong refactoring to compute this edf approximation. I feel that the subfunction swe_hyptest would need to be refactored in order to compute the SwE within it. However, I believe other tasks are more urgent before doing that.

### DIFF
--- a/swe_cp_WB.m
+++ b/swe_cp_WB.m
@@ -1413,7 +1413,6 @@ else % ".mat" format
         else
           hyptest=swe_hyptest(SwE, score, CrS, cCovBc, Cov_vis, dofMat);
         end
-        hyptest=swe_hyptest(SwE, score, CrS, cCovBc, Cov_vis, dofMat);
         p = hyptest.positive.p;
         edf = hyptest.positive.edf;
         clear CovcCovBc cCovBc
@@ -1911,11 +1910,8 @@ for b = 1:WB.nB
       % Calculate TFCE uncorrected p image.
       if TFCE    
 
-	% Obtain P values.
-	hyptest=swe_hyptest(SwE, score, blksz, cCovBc, Cov_vis, dofMat);
-	
-	% Current XYZ indices
-	currXYZ = XYZ(1:3, index);
+        % Current XYZ indices
+        currXYZ = XYZ(1:3, index);
 	  
         % T test already converted to Z
         if strcmp(WB.stat, 'T')
--- a/swe_cp_WB.m
+++ b/swe_cp_WB.m
@@ -320,6 +320,9 @@ end
 
 %-preprocessing for the modified SwE
 if isfield(SwE.type,'modified')
+  if dof_type == 1
+    error('degrees of freedom type still not implemented for the modified SwE and the WB')
+  end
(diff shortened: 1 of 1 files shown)


nicholst:
Quick question: Just want to confirm that DF Approximation I is a
recommended option? And in which settings?

And, I understand you're saying DF Approximation I won't be available for
modified SwE due to incompatibility with swe_hyptest?

BryanGuillaume:
Approximation I is the safest option for the classic SwE when using a parametric analysis, but it is not optimal because it yields a conservative control of the FPR. That is the only one implemented for the classic SwE in the fsl version. At the time, we decided to simplify things by offering only a single approximation in the fsl version i.e. Approx III, which is clearly the best choice for the modified SwE and is equivalent to Approx I when the classic SwE is used.

In the SPM toolbox, Approx II and Approx III are never selectable by the user anyway when a classic SwE is used (only naïve or Approx I). Only the modified SwE had all 4 options selectable (naive, Approx I, II & III). All of this have been working for the parametric test. However, Approx I has never been coded for the WB. 

This MR is thus implementing it for the WB, both for the classic and modified SwE. The issue was that the refactoring done using the subfunction `swe_hyptest` is not really appropriate for implementing Approx I because it needs some temporary variables that are not available when `swe_hyptest` is called. Thus, for now, I simply did it outside of this function for both the classic and the modified SwE. 

The main issue I have is that I could quickly validate the code only for the classic SwE by comparing with the FSL results. Nevertheless, as Approx I for the modified SwE is not coded in the fsl version, I could not benchmark it yet. Here, my main goal was more to validate the FSL code anyway due to the release deadline of the fsl version.

nicholst:
OK, understood.  Thank you!

BryanGuillaume:
To avoid losing too much time testing whether this is correct for the modified SwE, I have decided to disable Approx I for the WB modified SwE. It will still be available for the parametric tests and the WB classic SwE, which have been already tested. As I have already said before, this approximation is anyway never better than Approx. III, which is already implemented for the modified SwE. 

I will wait for the Travis checks to pass and then I will merge this PR because I would like to test if the changes in PR #138 are valid with the Approx I, WB and classic SwE options.

PR #138, opened 2019-08-23
Add GIfTI format

BryanGuillaume:
The goal of this PR is to allow the toolbox to analyze GIfTI files. To enable that, many changes have to be done (58 commits and 17 files changed so far...). 

The main change is the use of new SPM12 functions that are able to read/write GIfTI files like, for example, spm_data_read instead of spm_get_data. For this reason, I think the toolbox will not work anymore with SPM 8. To keep the compatibility, we could

1. use our own read/write GIfTI functions, independent of SPM12. The issue is that we may miss some updates of the SPM12 functions
2. use some form of wrapper function with an if-else condition that would detect if the SPM12 functions exist and use the SPM12 read/write GIfTI functions if available. That would allow being up-to-date with SPM if there are some changes but would prevent to read GIfTI files for SPM8 users

@nicholst, I am actually wondering how important is the compatibility with SPM8 as this becomes quite old now and if we should maintain this compatibility. 

Another important change is the loop over chunks of data. Before, the toolbox was looping over slices of volume data and eventually over blocks of voxels within these slices. Obviously, this does not work with surface data and the toolbox now loop over chunks of data like SPM12 is now doing.

I am currently testing if these changes did not break anything for volumetric data and have caught already a few bugs (there are still a few more options that I want to check). Nevertheless, I will also need to benchmark on GIfTI data. @nicholst, do you have some GIfTI files that I could use for this? 

This work is still in progress as I have to make sure that there is no bug left (I am also curious to see if the automatic tests with Travis will pass) and we might need to handle the issue of compatibility with SPM8.


BryanGuillaume:
All the WB tests passed, but not the one for the parametric test. That is strange as I could validate them on my computer. I will try to check why this is the case here.

BryanGuillaume:
After checking why Travis checks failed for the parametric tests, it appears that this is due to a change in the way the SwE is computed for the modified version:

- Before this PR, the estimate of Cov(Y) was estimated and saved as images in the loop over slices.  The SwE was computed after the loop over slices by loading the estimate of Cov(Y) from the saved images. The reason for the latter was that I was investigating the possibility to smooth the estimate of Cov(Y) before computing the SwE, which appears to be a bad idea for, at least, parametric tests.
- After this PR, the SwE is estimated directly after Cov(Y) in the loop over chunks. This avoids the need to read the Cov(Y) images.

Because the values of Cov(Y) are saved in single float (32 bits) format and the variable in memory are per default in double format (i.e. 64 bits), there are some discrepancies appearing when computing the SwE, which propagates the other statistics derived from the SwE. I believe that this is the only reason why the tests failed. Here, I see 2 options:

1. Use the new results as new ground truth
2. Allow some tolerance when comparing the results.

@nicholst, I am not sure if you have a preference regarding this? For the option with tolerance, the issue is to find a good tolerance value, particularly because the discrepancy tends to increase for estimates based on the SwE estimates. For example, for the SwE, the discrepencies seems to be around 10^-12 while for the edf, I could observe a discrepency value around 2 10^-6 (for a estimate value of 11). For this PR, I believe that using a relative tolerance of 10^-6 would be necessary for all the Travis tests to pass.

nicholst:
@BryanGuillaume I don't have strong feelings... we probably shouldn't trust our final results much beyond single precision floating point ("float"), but it would seem silly to save intermediary results in anything less double precision ("double").

For the record, SPM uses:
* `uint8` for masks
* `float32` for beta, contrast and statistic images
* `float64` for ResMS and RPV

I'm not sure exactly of the logic for this, except that ResMS and RPV enter into final calculations as divisors, and hence justify greater precision.

Following the lead from SPM, maybe saving the covariance in double is a good idea... on the other hand, we have a covariance *matrix* to save, which then doubles the space.

I'm happy either way... just make a decision.

BryanGuillaume:
Just for information, I have decided to update the ground truth of the test data instead of using a tolerance. The reason for this choice is that I feel that it is useful to detect any form of discrepancies, even if there are tiny and non-problematic.  Now, all the checks have passed.

BryanGuillaume:
Only for information, I have done additional tests on my computer to check if NIfTI inputs work properly and I can confirm that everything seems to work as intended. I have also started to test GIfTI files for parametric analysis. So far, the analyses seem to produce results without obvious error. Here below a screenshot of an analysis result:

Screenshot 2019-09-02 at 18 09 28

I created my GIfTI files using a spherical atlas, which might not be a good idea for observing the results as it is difficult to see what the significant vertices are. Thus, I might try to find a "less inflated atlas", at least for testing. 

An issue that I could detect so far is that, while I can rotate the image to observe the results around the sphere, I can also rotate the result table, which is annoying. I may need to check if this is the case with a classic SPM analysis. I will also test if  non-paramtric analysis are working properly.

nicholst:
On the table rotation problem - send a quick email to Guillaume (and/or tag
him here) and see if he has any ideas. .
-- 
Thomas Nichols, PhD

BryanGuillaume:
Regarding the rotation of other objects than the surface image, I can confirm this is also the case for classic SPM analysis. For example,

Screenshot 2019-09-03 at 11 28 12

Thus, I believe that it is out of the scope of our toolbox to deal with this. Note that just changing the contrast or significance level reset all the objects in their normal position. Thus, I guess that is an acceptable annoyance for now.

nicholst:
Tagging @gllmflndn to show him the  '[rotated table](https://github.com/NISOx-BDI/SwE-toolbox/pull/138#issuecomment-527383586)' bug (in SPM GIfTI results), in case he wasn't aware of it.

BryanGuillaume:
I have been testing and debugging WB analyses for GIfTI files. After a few bug fixes, it seems now to produce valid outputs. Nevertheless, on the contrast tested, I am getting several equivalent scores with `Inf` values, which appear to be missing from the display (see the holes in the second thresholded image below). This issue might not be related to this PR and may have been present before, but the equivalent scores were not high enough to yield `Inf` values. In particular, this does not seem to arise for parametric tests (see firts image below). Thus the solution to this issue is likely to already be in the code related to the parametric tests. I will look into this now. Here below, a comparison of results WB vs. param.

### Parametric test (5% FDR)
Screenshot 2019-09-05 at 14 58 52

### WB (999 bootstraps, 5% FDR)
Screenshot 2019-09-05 at 14 40 57

nicholst:
OK... thanks for the update.

gllmflndn:
It's a feature :yum: I never really worried about that as you should just not rotate these axes! The fix is to use `setAllowAxesRotate`:
https://github.com/spm/spm12/blob/r7487/spm_mesh_render.m#L147-L150
I don't remember why I disabled it but this could be revisited. When in rotate mode, you don't have access to the callbacks on the design matrix and the results table; that's why there is a Rotate switch in the contextual menu of the patch.

Initially, I thought you were commenting on the SPM{F_3,308} that is flying around when you rotate the mesh - I quite like it...

Note that a limitation of the implementation in SPM12 is that it requires a specific form for the GIfTI files (data stored in a separate binary file and link to the geometry of the mesh through the metadata).

nicholst:
Thanks Guillaume! (This GIfTI work is a building block on the way to
support CIFTI).

BryanGuillaume:
@gllmflndn, thanks for your explanation about the rotation of the table and to tell us how to fix it. 

Regarding the observed Inf values for the X-scores, the previous commit seems to fix it. Now, the X-scores are computed using

`swe_invNcdf(0.5 * p).^2` instead of `spm_invXcdf(1 - p, 1)`

where `p` is the p-value. This strategy is much better for small p-values and has been already utilised for parametric tests. As a proof that this is working, here below the same analysis as above but after this commit:

Screenshot 2019-09-05 at 19 33 49


The Inf values and the "holes" are gone!

nicholst:
Super! Thanks Bryan!

gllmflndn:
`spm_invNcdf` now also uses `erfcinv`. For `spm_invXcdf`, I've been comparing with your approach and [chi2inv](https://www.mathworks.com/help/stats/chi2inv.html):

invXcdf

Maybe more precise estimates for any degrees of freedom can be obtained by improving `spm_Gcdf`, i.e. using the `tail` option in `gammainc`?

BryanGuillaume:
Thanks @gllmflndn for this plot. It shows well when `spm_invXcdf` and `chi2inv` struggles.  To overcome this issue for any degrees of freedom, the functions must work with `p` instead of `1-p` and, as you said, it seems that the only way would be to use the 'upper' tail option of `gammainc` and add a tail option in the `spm` functions which depend on it (i.e. `spm_Gcdf`, `spm_invXcdf`,...). 

The question is to know whether this is really needed in practice. For the SwE toolbox, we are only converting score in X-score with 1 degree of freedom. Thus, the current strategy based on `spm_invNcdf` is good enough for us. Nevertheless, if SPM or other SPM toolboxes use conversion  in X-score with more than one degree of freedom, it might be useful to add a tail option.

gllmflndn:
@BryanGuillaume: Thanks. I did add a `tail` option to `spm_Gcdf` a few months ago, next step would be to have `spm_invGcdf` use it when required. As you say, it's probably a rare use case so I'm going to leave it as is for now but this can be revisited if and when this discussion comes up again.

BryanGuillaume:
As a last set of checks for this PR, I have run a comparison between the SwE toolbox and a classic SPM analysis on the contrast comparing one category (FACE) vs the average of three other categories (BODY, PLACE & TOOL). For the classic SPM, I have computed the contrast image prior using a one-sample t-test. For the SwE toolbox, I have used all the contrast images, selected the modified SwE with 2 homogeneous groups (the only reason of 2 instead of 1 was just for testing purposes; in a real data analysis I might have picked 1 homogeneous group). The results seems consistent. For example, here below some results for the negative effect:

### SPM negative effect
spmNeg

### SwE negative effect (parametric test)
sweNeg

### SwE negative effect (voxel-wise non-parametric test)
Screenshot 2019-09-10 at 12 55 57


### SwE negative effect (cluster-wise non-parametric test)
Screenshot 2019-09-10 at 12 56 21

I think I have benchmarked enough this PR to validate it and, as soon as the Travis test data is updated to reflect the change in the computation of X-scores for F-contrasts (see comments above), I will merge this PR.

BryanGuillaume:
All tests passed! I am merging.

PR #141, opened 2019-09-25
Add CIfTI support

BryanGuillaume:
This is still a work in progress. The goal of this PR is to offer the possibility to analyse CIfTI inputs (currently only `dtseries.nii` and `dtscalar.nii`). So far, this has been done for parametric analyses only and I am currently checking if there is no bug left for this type of analyses. After validating this, I will implement CIfTI support for WB analyses as well. Below, a brief summary of what this PR is actually doing:

- Creation of a `@cifti` directory behaving like the `@nifti` directory in SPM but tuned for CIfTI files
- All the `spm` functions reading/writing file headers and data have been replaced by `swe` functions that either call the corresponding spm functions if the file is not CIfTI or specifically deal with CIfTI files
- Creation of a specific cluster function for CIfTI files: `swe_cifti_clusters`. It works similarly like `spm_clusters` and `spm_mesh_clusters` but for CIfTI inputs.
- Creation of a specific max statistic function for CIfTI files: `swe_cifti_max`. It works similarly like `spm_max` and `spm_mesh_max` but for CIfTI inputs.
- Now the user has an option to specify the geometry files of the surface brain structures in the CIfTI files. I will show a screenshot later showing this. 
- Creation of a function `swe_read_cifti_info` that reads some information from the embedded xml file of CIfTI inputs. Currently, it extracts only what the toolbox needs, but this can be improved if needed.
- A specific CIfTI display or results has been created (I will show some screenshots later with more detailled explanations)


BryanGuillaume:
Here below a screenshot of the model specification with the new field allowing the specification of geometry files for the surface brain structures in the CIfTI files. Because the toolbox cannot know in advance how many surface brain structures are in the CIfTI files, the user needs to add new brain structures in a similar fashion than adding covariates. The user also needs to provide the label of each surface brain structure as it is specified in the CIfTI files in order to match each geometry file to the good brain structure. The toolbox does some check when it runs the model specification module to make sure all the needed geometry files have been specified.

Screenshot 2019-09-25 at 12 37 49

BryanGuillaume:
Here below a screenshot of the CIfTI display of results. The current choice is to display the results for all the brain structures. It is important to note that all the volume structures has been merged and display as a single volume with the label `V` (for the example, they were actually 19 volume brain structures in the CIFTI files). For the surface brain structures, they are not merged and displayed separately using the labels `S1`, `S2`,...  The table uses the same sorting of clusters than a typical SPM analysis i.e. it orders the clusters using the max statistic of each cluster. To be able to know in which brain structure they are, I have added  the brain structure label for each cluster. @nicholst, do you think this solution of displaying the cifti results is ok? Or do you have another option in mind? Please note that more change may be needed for the WB display of results (e.g., other statistics may be needed).


Screenshot 2019-09-25 at 12 46 11

nicholst:
@BryanGuillaume This display is better than I could have possibly hoped for!

Question:  Do you assume/require that you have 'brainordinate' data?  Will it work with only surfaces?

Also, do you require geometry files?  That's fine, if so, but just was thinking that, for voxel-only, it isn't needed (though makes visualisation basically impossible).

BryanGuillaume:
The toolbox should, in theory, works with any numbers of surfaces and volumes including 0. I would, however, need some specific test data to verify that there are no bugs when one of the brain structure types is not present (e.g., attempting to access an empty object which would result in an error), but the code has been written with this in mind. Thus I am not expecting much bugs related to this.

In the specification module, only the geometry files for the surface structures are required. If the user fails to specify them, the toolbox should throw an error when running this module. For the volumes, the geometry info is already in the CIfTI files and the toolbox is extracting this using a new function swe_read_cifti_info. 

I have noticed that Travis does not like something in this PR as octave throw the following error: `cell cannot be indexed with .`. @TomMaullin or @nicholst, have you run into this type of errors in octave before (that does not seem to exist in MATLAB).

BryanGuillaume:
For information, the bug `cell cannot be indexed with .` in octave has been fixed.

nicholst:
@BryanGuillaume Please, on rescomp, `XXXXXX/T1w/fsaverage_LR32k` for `*dscalar.nii` images that contain just two surfaces but no volume data.  I honestly don't know what the intensity values are, but should suffice for testing.

nicholst:
Update on the meaning of the test images.

I have pointed you to some L+R surface (no volume) images, 
```
XXXXXX.midthickness_MSMAll_va.32k_fs_LR.dscalar.nii
XXXXXX.midthickness_MSMAll_va_norm.32k_fs_LR.dscalar.nii
```

I can now provide an interpretation of these images thanks to FMRIB's @MichielCottaar 

> "va" stands for "vertex area". These files contain the size of the vertices at the mid-cortical surface (i.e., halfway between the white/gray matter boundary and the pial surface). With the size of the vertices, I mean that the size of each triangle in the mesh is calculated. The size of the vertices is then defined as the sum of the sizes of the triangles touching that vertex divided by 3 (because each triangle touches 3 vertices). 
> 
> In the va_norm all the vertex areas have been normalised with a scalar, so that the total surface area equals 59412. This means that by taking this value, you can compare the relative surface area taken up by a vertex between subjects (i.e., corrected for differences in the total cortex area).

Thank you Michiel!

BryanGuillaume:
### Update regarding the use of CIFTI files with multiple images (also referred to as frames below)

Several changes have been done to the code to allow the use of CIfTI files containing several images. More precisely, when selecting the scans, a frame filter (the same as for NIfTI files) can be used to select which image to select. The toolbox is also able to detect on the fly if the required frame numbers exist in the CIfTI files. Here below, an example of this feature:
Screenshot 2019-10-01 at 15 44 00

To check if this seems to work properly, I have run an analysis selecting the same 4 tasks as above in the multiple-image CIfTI files MSMAll_All_task_maps_XXXXXX.dtseries.nii. While they do not contain contrast values like in the analysis above but t-scores, the results of this analysis are very similar than those obtained previously (see the 2 screenshots below), indicating that this seems to work correctly. 

### Analysis based on the multiple t-score image CIfTI files
Screenshot 2019-10-02 at 12 35 29

### Analysis based on the single contrast image CIfTI files
Screenshot 2019-10-02 at 12 35 01

BryanGuillaume:
### Update regarding the use of CIFTI files with surface data only (i.e no volume data)

After a small bug fix, this seems to work as well. For example, here below the result of an analysis on the XXXXXX.midthickness_MSMAll_va_norm.32k_fs_LR.dscalar.nii files that contain only surface data:

Screenshot 2019-10-02 at 17 30 15
  
Note that, due to the way I design this analysis, there should not be any effect detected, which is the case if we do a multiple comparison correction (see, e.g., FDR-corrected p-values in the table)

nicholst:
Super!  Great to see.

Are we getting close to send to beta testers?

BryanGuillaume:
I believe the toolbox is ready for parametric analyses of CIfTI files. However, there is still some work left for WB analyses, but I think a beta could be ready at the beginning of next week.

BryanGuillaume:
@nicholst, I have modified the code to enable WB analyses with CIfTI inputs. For voxel-wise analyses, this seems to work as intended. As discussed previously, the difficulty is for cluster-wise inference. Currently, I have implemented 3 strategies:

1. Use of the number of voxels/vertices in each cluster (classic approach)
2. Computation of the null distribution of (cluster sizes).^(1/d) for each type of brain structures based on nB bootstraps (d is 2 for surfaces ad 3 for volumes). Then, for each type of brain structures, normalisation of (original cluster sizes).^(1/d) by removing the median of the null distribution and dividing by the IQR of the null distribution. These normalised cluster sizes are then used to compute the FWER p-values (normalised approach 1).
3. Computation of the null distribution of (cluster sizes).^(1/d) for each type of brain structures based on nB bootstraps (d is 2 for surfaces ad 3 for volumes). Then, for each type of brain structures, normalisation of (original cluster sizes).^(1/d) by dividing by the median of the null distribution. These normalised cluster sizes are then used to compute the FWER p-values  (normalised approach 2).

From an experiment with 999 bootstraps, it seems that the null distribution of cluster sizes even after transformation (*).^(1/d) is highly skewed with a median of 1 for volume and a median of sqrt(2) for surfaces (see image below). They actually looks like geometric distributions. Thus I am not sure that the normalisation are actually appropriate.

### Null distributions of (cluster sizes.^1/d)
Screenshot 2019-10-08 at 12 45 07

The difference on -log10(cluster-FWE p-values) can be seen on the image below:

Screenshot 2019-10-08 at 11 39 51

The normalised approach 1 tended to improve the p-values of some clusters in the volume while penalising some clusters in the surfaces. On the contrary, The normalised approach 2 tended to improve the p-values of some clusters in the surfaces while penalising some clusters in the volume.

Overall, I am not sure that the suggested normalisation are doing a good job. @nicholst, what do you think about these results?

nicholst:
Can you explore with Box Cox?   But, also, can you also 'explore the
parameter space' more, it seems like the data should be smoother.... i.e.
generate more bigger clusters...

BryanGuillaume:
For information, this skewness towards very small clusters seems to be also present for the original clusters:

- For the volume
  - 35 % of original clusters have size 1 (51 % of null clusters have size 1)
  - 12 % of original clusters have size 2 (21 % of null clusters have size 1)

- For the surfaces
  - 17 % of original clusters have size 1 (40 % of null clusters have size 1)
  - 14 % of original clusters have size 2 (21 % of null clusters have size 1)

BryanGuillaume:
I have experimented with BoxCox transformation, but I do not think the results are better. Here below a comparison of -log10(clusterwise-FWER p-values) obtained following different transformation with

- classic: number of elements in each clusters
- normalised 1: `((original cluster size)^(1/d) - median((null cluster size)^(1/d))). / iqr((null cluster size)^(1/d)))`
- normalised 2: `(original cluster size)^(1/d) ./ median((null cluster size)^(1/d))`
- normalised 3: `(boxcox(lambda_null_surf/vol, original cluster size) - median(boxcox(lambda_null_surf/vol, null cluster size))). / iqr(boxcox(lambda_null_surf/vol, null cluster size))`
- normalised 4: `((original cluster size)^(1/d) - mean((null cluster size)^(1/d))). / std((null cluster size)^(1/d)))`
- normalised 5: `(boxcox(lambda_null_surf/vol, original cluster size) - mean(boxcox(lambda_null_surf/vol, null cluster size))). / std(boxcox(lambda_null_surf/vol, null cluster size))`
Screenshot 2019-10-08 at 18 08 48

The best trade-off might be achieved with the version "normalised 4" that replaces the median and the IQR by the mean and the standard deviation, respectively, but this may be a very subjective assessment and may not be the case in other scenarios. @nicholst, do you know what would be  a good criterion to assess this more objectively?

I am also wondering if the max statistic null distributions could be better behaved for a normalisation as they should be less skewed towards small clusters. I will try to investigate this.

BryanGuillaume:
@nicholst, I have applied the same transformations as above but based on the null distribution of the max cluster sizes instead of the cluster sizes. The results seem better behaved as can been seen below. Typically, some volume p-values seem to systematically improve while some surface p-values are penalised but in a more reasonable way than the results above. Also, all the transformation seems to yield very similar results. Maybe a good way to check if we are doing a good job would be to compare the cdf of each volume transformation vs surface transformation. Do you have any opinion about this?

Screenshot 2019-10-09 at 12 52 23

BryanGuillaume:
After reviewing my code, I found a bug in the computation of the normalised version 4 based on the null distribution of cluster sizes. Here below the updated figure. The results of this version are now similar to those of the normalised version 1, which makes more sense. I will edit the comment above to warn about this error.

Screenshot 2019-10-09 at 18 04 28

BryanGuillaume:
As additional plots to compare the normalisations procedure, here below are a comparison between the empirical cdf of each normalised distribution.

### Normalisation based on the null cluster sizes
Screenshot 2019-10-09 at 18 29 13
### Normalisation based on the null max cluster sizes
Screenshot 2019-10-09 at 18 32 06

BryanGuillaume:
After investigating different normalisation strategies, I have finally implemented two normalisations based on the box-cox transformation. The first version correspond to `(boxcox(lambda_null_surf/vol, original cluster size) - mean(boxcox(lambda_null_surf/vol, null cluster size))). / std(boxcox(lambda_null_surf/vol, null cluster size))` and the second version correspond to `(boxcox(lambda_null_surf/vol, original cluster size) - median(boxcox(lambda_null_surf/vol, null cluster size))). / 2*(Q3(boxcox(lambda_null_surf/vol, null cluster size)) - median(boxcox(lambda_null_surf/vol, null cluster size)))`. It is important to note that these two normalisations, particularly the first one, seemed to be appropriate when the data was smoothed but not really when the data was not smoothed. It seems that the main reason for this is that the distribution of the cluster sizes under H0 tends to be highly skewed towards 1 for rough data while this is much less the case for smoothed data.  

I have also modified the display of results. Now, for an FWER-clusterwise inference with CIfTI, the user can select between an inference based on the classic cluster sizes (i.e number of vertices or voxels), the first normalisation (BC1) and the second normalisation (BC2). I have also modified the cluster-level values in the table as it can be observed below. I have also added the Box-cox parameters in the footer section.

I have run a WB analysis with the same contrast of interest as above on the unsmoothed data and here below are the results of cluster-wise inference based on the 3 cluster size type available. we observe that the normalisation typically improves the p-values for the surface clusters but penalises those of the volume clusters (there is even no volume cluster surviving for BC2...).  

Activation

To look further at all the range of p-values, here below are the -log10 p-p plots for activation and deactivation.  We clearly see that the normalisation tends to favour the clusters in the surface data and tends to penalise the clusters in the volume data. I am not convinced that this is the effect that we are looking for. However, we may expect to see a different behaviour for smoother data.
 
Screenshot 2019-10-17 at 20 41 05
Screenshot 2019-10-17 at 20 41 09
Screenshot 2019-10-17 at 20 41 02
Screenshot 2019-10-17 at 20 41 00

nicholst:
One quick comment:

Please scale IQR by 1/(norminv(0.75)-norminv(0.25)) = 1/1.348979500392164 so that the normalisations are equivalent and have comparable units.

BryanGuillaume:
I have been doing some further checks, fixing a few remaining bugs, modifying a part of the result report for CiFTI data and modifying some help text related to CIfTI data.

I believe I am now ready to merge this PR and make a beta release with all the last PRs. I will do so as soon as Travis confirms that all checks pass.

BryanGuillaume:
The Travis tests related to the cluster-wise analyses have failed. This is due to the addition of the cluster extent images that were not saved previously (see this [commit](https://github.com/NISOx-BDI/SwE-toolbox/pull/141/commits/08d717cff7b0d495ac6196543dc18076861b2b97)). I will add these files to the test data repo.

BryanGuillaume:
After several attempts with Travis, all Travis checks have finally passed. I will then merge this PR

PR #145, opened 2019-10-28
Fix several CIfTI file issues 

BryanGuillaume:
The goal of this PR is to fix some issues observed with CIfTI files. Here below a summary of the fixes:

1. The XML files of the CIfTI inputs are now always copied in the CIfTI outputs. Before this PR, that was not the case.
2. The XML files of the CIfTI inputs with more than one contrast are modified in the outputs to set the number of time series to 1 (i.e. NumberOfSeriesPoints="1")
3. the nan values in the outputs have been replaced by zero values to avoid issues with CIFTI files

### DIFF
--- a/@swe_cifti/private/write_hdr_raw.m
+++ b/@swe_cifti/private/write_hdr_raw.m
@@ -83,6 +83,15 @@ if sts
             sts = false;
         end
     end
+
+    if isfield(hdr, 'ext')
+      % write the XML file
+      fwrite(fp, [1 0 0 0], 'uint8');
+      len1 = fwrite(fp, hdr.ext.esize, 'int32');
+      len2 = fwrite(fp, hdr.ext.ecode, 'int32');
+      len3 = fwrite(fp, hdr.ext.edata, 'char');
+      len4 = fwrite(fp, zeros(1, hdr.ext.esize - 8 - numel(hdr.ext.edata)), 'uint8');
+    end
     fclose(fp);
 end
 
--- a/@swe_cifti/subsasgn.m
+++ b/@swe_cifti/subsasgn.m
@@ -381,8 +381,13 @@ if isa(val,'file_array')
                 obj.hdr.magic = ['n+1' char(0)];
             end
         case {'.dtseries.nii','.dtscalar.nii'}
-            hdr = read_hdr_raw(sval.fname);
-            val.offset    = max(sval.offset, 544 + hdr.ext.esize);
+            if isfield(obj.hdr,'ext')
+              offset = 544 + obj.hdr.ext.esize;
+            else
+              hdr = read_hdr_raw(sval.fname);
+              offset = 544 + hdr.ext.esize;
+            end
+            val.offset    = max(sval.offset, offset);
             obj.hdr.magic = ['n+2' char(0) sprintf('\r\n\032\n')];
         otherwise
             error(['Unknown filename extension (' suf ').']);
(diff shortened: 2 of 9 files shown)


BryanGuillaume:
I will wait for Travis to validate these changes. If Travis validates them, I will directly merge in order to make a new candidate release 2.2.0.rc2 as soon as possible.

BryanGuillaume:
For information, I have been testing it on these changes on CIFTI parametric analyses, but I have not had the time to do the same on WB analyses.

nicholst:
Thank you!

BryanGuillaume:
mmm, I have some weird errors in Travis. For example:

```
Testing file: swe_vox_Tstat_c01.nii         
322error: verifyMapsUnchanged: mx_el_ne: nonconformant arguments (op1 is 153594x1, op2 is 51560x1)
```

This might be due to the fact that we are now using 0s instead of nans and thus, some file comparison are failing. I will investigate this in more details

nicholst:
Awesome... only potential scary thing to check is about an edit to the XML
that changes its length.  Do you have example inputs with
2-digit NumberOfSeriesPoints, which you will then reduce to a 1-digit (i.e.
"1") value?  Will that mess anything up?

ALSO, are you aware that NIFTI vox_offset *must* be aligned to 16-byte word
boundaries?  See here
https://nifti.nimh.nih.gov/nifti-1/documentation/nifti1fields/nifti1fields_pages/vox_offset.html


It may be that CIFTI doesn't worry about this, but since SPM uses memory
mapping maybe it matters?  And, then, I won't know what you would do except
add spaces at the end of the XML maybe?

Anyway... just a gotcha to think about...

BryanGuillaume:
Yes, that is a good point that made me a bit worried. Nevertheless, one change of this PR in @swe_cifti/private/write_hdr_raw.m consists of writing the XML file to the CIFTI file by adding 0s at the end (see [these lines](https://github.com/NISOx-BDI/SwE-toolbox/blob/a144fd93e1681e88e88c086fcd609970803426f9/%40swe_cifti/private/write_hdr_raw.m#L87-L94)). This should do a correct job taking into account the fact that when reading the XML files, the trailing zeros are actually set to empty characters (see [here](https://github.com/NISOx-BDI/SwE-toolbox/blob/b5fef5b19493ae11f2760bdc6336a2499108e588/%40swe_cifti/private/read_hdr_raw.m#L153)).

BryanGuillaume:
While waiting for Travis, I have run a quick WB analysis with 99 bootstraps and everything seems fine ate first glance.

BryanGuillaume:
All Travis tests have finally passed. I am merging this PR

nicholst:
OK... extra NULL char's should be safe.  But, out of curiosity, in all of
the offsets you've seen in CIFTI's flavor of NIFTI, are they all multiples
of 16?  If so, it would seem important to ensure the continue to be so.

BryanGuillaume:
Just replied on this topic to one of your emails. I am repeating the content below. Yes, I had three flavors of CIFTI outputs and all three give me the following for the offset:

vox_offset: 631936 = 16 * 39496
vox_offset: 657728 = 16 * 41108
vox_offset: 343232 = 16 * 21452

Thus, yes they are multiples of 16.

nicholst:
Super!  And, since you're at most shrinking the length of the XML by one or
two bytes and add NULL padding at the end, you're never going to change the
actual value of the offset on writing out new CIFTIs, right?

PR #147, opened 2019-11-01
Keep only one "NamedMap" in the XML of the SwE dscalar.nii outputs

BryanGuillaume:
The goal of this PR is to keep only one "NamedMap" in the XML extension of the toolbox dscalar.nii outputs.

The code change seems to work as can be seen in the screenshot below. However, the "NamedMap" that is kept (i.e. WM_15_BODY) corresponds to the "NamedMap" of the slice selected for the first dscalar.nii input. As, in an analysis, other slices could be selected as well (in this specific analysis, I was using 4 different slices), that might not be so great. Another solution would be to write a specific name for the swE output like "SWE_OUTPUT" to avoid confusion. @nicholst, do you think this is an issue worth to be worried about? 

Screenshot 2019-10-31 at 20 30 42

### DIFF
--- a/swe_data_hdr_write.m
+++ b/swe_data_hdr_write.m
@@ -57,6 +57,21 @@ function V = swe_data_hdr_write(fname, DIM, M, descrip, metadata, varargin)
         xml(ind+1) = [];
       end
     end
+
+    % modify the xml if there is more than one <NamedMap>
+    ind = strfind(xml, '<NamedMap>');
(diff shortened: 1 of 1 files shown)


nicholst:
I won't worry about it.  In all of the use cases I can think about, the
same map will be pulled from each file.

Thank you!

BryanGuillaume:
Merging this PR as this seems to work properly

Issue #150, opened 2019-11-12
Replacing NaN values by zeros seems to have some unwanted side effects


BryanGuillaume:
Following the request to replace NaN values by zeros in the result images to prevent issues with other software (see this [commit](https://github.com/NISOx-BDI/SwE-toolbox/pull/145/commits/a144fd93e1681e88e88c086fcd609970803426f9)), it seems that there are some undesirable side effects. For example, the minimum number of degrees of freedom is now 0:

<img width="219" alt="Screenshot 2019-11-12 at 16 36 54" src="https://user-images.githubusercontent.com/4426961/68685596-a61fde00-056a-11ea-96c7-1b30a2294f97.png">

I suspect that this is due to the fact that the edf image is read entirely while it would be more appropriate to read only the in-mask data.  I will review all the possible side effects that this change could cause and try to fix them

BryanGuillaume:
I have detected two other bugs related to this:

1. The computation of the 5% FDR threshold in `swe_list.m` is including the zeros values, which is likely to bias the threshold (see [here](https://github.com/NISOx-BDI/SwE-toolbox/blob/a48d5e5b708a2bc451cd7530e0d14be80cdeae83/swe_list.m#L450-L458))

2. The extraction of edf summary statistics (see [here](https://github.com/NISOx-BDI/SwE-toolbox/blob/a48d5e5b708a2bc451cd7530e0d14be80cdeae83/swe_list.m#L537-L547)) does not work anymore. Indeed, the line `edf = swe_data_read(xSwE.Vedf);` loads the data as a 3D array  and, before, the line `edf(isnan(edf)) = []` was vectorising it providing that there were some NaN values. Now, because all NaN values have been removed, there is no vectorisation of `edf` anymore. That causes the extraction of the max/min/median statistics by using, for example, `max(edf)` to fail as they now yield a matrix instead of a scalar, causing the toolbox to throw an error when extracting these summary statistics.

I will look if there are other issues that may occur.

BryanGuillaume:
All the bugs observed due to replacing NaN values by zeros seems to have been fixed in PR #152. Thus, I am closing this issue.

Issue #148, opened 2019-11-02
Change CIFTI output naming to match PALM


nicholst:
PALM uses 'dpx' instead of 'vox' for the filename type when reading CIFTI; change to match PALM.

BryanGuillaume:
I agree that `vox` might no be appropriate for CIFTI outputs. However, I am not sure that `dpx` was really intended for CIFTI data as it seems to stands for "surface-data without specification" 

Indeed, from PALM documentation,

> This can be `vox` for voxelwise, `vtx` for vertexwise, `fac` for facewise, `dpx` for surface-data without specification (it can be vertexwise or facewise if the actual surface geometry wasn't supplied with the -s option), `dat` for any data that doesn't necessarily have a spatial meaning

As CIFTI can contain both surface and volume data, that does not seem a perfect naming. If I were to pick something from the list, I would use `dat`, but this is quite a subjective choice.

Anyway, if the only goal is to match PALM as best as we can, I can change it to `dpx`. On this note, I should also change output naming for GIFTI data, but I am not too sure how to detect if the data is vertexwise or facewise. Maybe, I should use `dpx` per default for GIFTI and CIFTI data.

nicholst:
I agree `dpx` doesn't really seem perfect, but at this point I'm happy to just go along with the PALM convention. (And then we don't have to worry about whether area has been considered, or whether voxels have also been considered.)

nicholst:
@BryanGuillaume I guess this can be closed now?

BryanGuillaume:
Yes, this has been handled in PR #151. Thus, I am closing this issue.

PR #151, opened 2019-11-15
Modify naming convention for CIfTI, GIfTI and .mat outputs

BryanGuillaume:
This PR is linked to issue #148. Its goal is to change the naming of CIfTI, GIfTI and .mat outputs to better match the naming convention of PALM. Below, a list of the changes:

- For CIfTI outputs: `swe_vox_...` -> `swe_dpx_...`
- For GIfTI outputs: `swe_vox_...` -> `swe_dpx_...`
- For .mat outputs: `swe_vox_...` -> `swe_dat_...`

No change has been done for NIfTI outputs. I have been testing the changes and this seems to work fine for CIfTI and GIfTI outputs. Nevertheless, I encountered some issues when I was testing for NIfTI outputs (see issue #150). Thus I will have to fix these first in another PR before being able to check further the code for NIfTI. 

Another issue that could be arising is the retro-compatibility for the display of old results that uses the old naming convention. Nevertheless, the SwE.mat  should contain the names with the old convention and thus this could not be an issue. I would need to test this more to check this.


BryanGuillaume:
I believe there are no obvious issues remaining from this PR as the Travis tests have passed. Thus, I am merging it.

PR #152, opened 2019-11-15
[WIP] Fix bugs related to zeroing nan values

BryanGuillaume:
The goal of this PR is to fix the bugs reported in issue  #150. To fix them, this PR saves the coordinates of all in-mask voxels in the variable `xSwE` and uses them to load only the in-mask values needed for computing the 5% FDR threshold and the summary statistics of the effective numbers of degrees of freedom. Before, the code was also loading the values outside the mask, which was fine when they were NaN values but not when they had been changed to zero values.

I am still looking if zeroing nan values could have caused more bugs in the code. If I can find them, I will fix them in this PR.


BryanGuillaume:
After further checks, I have caught and fixed one more potential bug coming from zeroing NaN values. I cannot see other obvious bugs related to this change and, thus, I will merge this PR.

Issue #154, opened 2020-01-29
The naive edf seems to be inflated for design matrix separable per visit  


BryanGuillaume:
In a current analysis with 78 subjects and a grand design matrix having an intercept for each repeated measure, the naive estimator of edf is giving me edf=154 for a contrast involving 2 of the repeated measures while the other estimators of edf give me 77, which seems accurate. The issue seems that the naive estimator of edf currently implemented assumes that the grand design matrix is only separable between groups of subjects and not between repeated measures and therefore computes the edf as the sum of the edf associated to each sub-design matrix. This is working correctly when the sub-design matrices contained different subjects, but not when they contain subjects in common.

Note that this does not seem to affect the other estimators of edf.

I am currently fixing this issue

BryanGuillaume:
This was fixed in PR #155

Issue #153, opened 2020-01-29
There is an error when trying to display a WB analysis with a naive estimation of the edf


BryanGuillaume:
Below, the error message:

> 29-Jan-2020 13:37:16 - Failed  'Results Display'
> Reference to non-existent field 'edf'.
> In file "/Users/bryanguillaume/Documents/GitHub/SwE-toolbox/swe_getSPM.m" (???), function "swe_getSPM" at line 1401.
> In file "/Users/bryanguillaume/Documents/GitHub/SwE-toolbox/swe_results_ui.m" (???), function "swe_results_ui" at line 0.
> In file "/Users/bryanguillaume/Documents/GitHub/SwE-toolbox/swe_run_results.m" (???), function "swe_run_results" at line 22.

I am looking into this error now

BryanGuillaume:
This was fixed in PR #155

PR #155, opened 2020-01-29
Fix several bugs related to the naive estimation of edf

BryanGuillaume:
The goal of the PR is to fix the 2 issues #153 and #154.

To fix issue #153, the piece of code throwing the error has been moved in an if/else statement to ensure it is only visited for parametric analyses.

To fix issue #154, the toolbox is now detecting which subjects are involved in the contrast tested and compute the naive edf based on these subjects. This seems to fix the observed issue.


BryanGuillaume:
After a few tests with the toolbox, this PR seems to fix the issues observed. Thus, I am merging it

PR #160, opened 2020-02-18
Fix a bug in the naive computation of the effective number of degrees of freedom

BryanGuillaume:
The goal of this PR is to fix a bug introduced in PR #155 regarding the computation of the naive estimation of the effective number of degrees of freedom.

The labels of the subjects that were involved in the contrast tested were not converted into indices prior to computing the naive edf. During the testing of the previous PR, the labels were actually matching the indices and the bug was not caught at that time.


BryanGuillaume:
After a few manual tests,  I think this PR corrects well the observed bug. Thus I am merging this PR.

PR #166, opened 2020-04-24
Modify WB cluster-wise inference

BryanGuillaume:
The goal of this PR is to modify the way cluster-wise inference is done with WB. More precisely, the goal is to use a normalised score of the cluster extent instead of the cluster extent itself as cluster statistic for all type of data. This normalised score was already implemented for CIfTI data and, as a reminder, is obtained as follows:

1. For each bootstrap, the cluster sizes of all surviving clusters under the null hypothesis are recorded. When possible, the cluster sizes are the surface areas of the clusters. Otherwise, this is simply the numbers of connected elements.
2. A Box-Cox transformation of all null cluster sizes is performed (not accounting for the cluster sizes originally obtained), yielding a Box-Cox `lambda` value. Note that, when both volume and surface data are present, the Box-Cox transformation is done separately in volume and surface, yielding two separate Box-Cox `lambda` values. 
3. The median (`Q2`) and upper-high IQR (`Q3-Q2`) of each Box-Cox transformations under H0 are also recorded.
4. The normalised score of a cluster of size `k_E` is then computed using the formula:
`k_Z = swe_invNcdf(0.75) * (boxcox(lambda, k_E) - Q2) / (Q3-Q2))`
where swe_invNcdf(0.75) is the third quartile of the standard normal distribution.
5. The resulting normalised scores are then used to make inference on the clusters

Another change introduced by this PR is that the Box-Cox normalisation based on the mean and std instead of the median and upper-high IQR that was previously available with CIfTI analysis has been removed. 

In addition, some refactoring has been made at several places in the code to make it cleaner.

I am still testing if everything works properly and fixing any bugs that I can detect. On this note, there may be cases where the number of null clusters is zero or very small. There might be also cases where the upper-high IQR is zero. These cases are likely to yield unsatisfactory normalised scores and, from a discussion with @nicholst, it seems better to revert back to use the unnormalised score as cluster statistic when this happens. I will update this PR to implement this.


BryanGuillaume:
Well, I have tried to follow some recommendations of "clean coding" by using `enumeration` in this PR instead of using char arrays or integers to describe categorical variables. Unfortunately, while this works well in Matlab, enumeration has not yet been implemented in Octave as can be seen in the following message

```
octave:1> enumeration
warning: the 'enumeration' function is not yet implemented in Octave

Please read <https://www.octave.org/missing.html> to learn how you can
contribute missing functionality.
error: 'enumeration' undefined near line 1 column 1
```

I will need to refactor this...

nicholst:
Hi @BryanGuillaume, This all looks great!  I'm so sorry about the lack of enumeration support in Octave.  I fear there's nothing that can be done about it... I'd refactor this PR to leave it out, but save the changes in another branch so we can maybe apply them later when Octave support is there.

Let me know if you need me to review anything else in detail.

BryanGuillaume:
As an update, I have fixed a several issues in a series of additional commits. In summary,

- I first replaced the enumeration by a class in the "new" Matlab OOP syntax (i.e. using `classdef`). Unfortunately, while Octave can use these classes to create objects and works with them, it cannot save these objects on disk. Thus, I have refactored a second time the class using the "old" Matlab syntax (i.e. using directories `@className`) and this seems to work properly now.
- I have deleted an enumeration class for cluster statistic types that was not needed anymore 
- I have replaced Matlab's `boxcox` function by two custom made functions. The reason for this was that the `boxcox` function is in the financial toolbox that the user may not have (as an example, Octave did not have it for the tests).
- I have also forced the boxcox transformation of the value `0` to be `-Inf` to avoid issues when transforming some max cluster size of 0 (for bootstraps characterised with no surviving clusters)
- Fix the misnaming of some variables 
- Fix some bugs specific to `.mat` inputs related to missing variables or a function acting on an empty cell array (and throwing an error)

All the tests related to non-clusterwise inferences are now passing. The tests related to clusterwise inferences are still failing, but this is due to the fact that we now have files related to the Box-Cox normalised cluster statistics using the naming pattern `clusternorm`. Obviously, they currently not exist in the ground truth of these tests. I will update this next.

BryanGuillaume:
I have fixed a few additional bugs that I caught while creating a new set of test data.  I have still to fix bug #168, implement a more complete handling of "bad" Box-Cox normalisations (currently, only Box-Cox normalisations with a upper-high IQR of 0 is considered "bad", which may not be enough; however I am considering to do this later in another PR as I am already handling the most critical case of "bad" normalisation), create test data for CIfTI inputs and finish benchmarking the code on this new set of test data to catch any remaining bugs.  Hopefully, after all of this, I will be able to finalise this PR.

nicholst:
Let me know if this is getting close...

BryanGuillaume:
After a few bug fixes, I can now run tests in Octave on my local machine on NIfTI (or hdr/img), GIfTI, CIfTI and .mat format (before, only test on NIfTI and .mat format were available). In parallel, I have also created a new set of test data (reduced in size compare to the old test data) for all 4 formats. For each format, the input data contains the same values. Thus, I will now check if the results obtained from the tests across each format are consistent. That should be the case for voxel/vertex-wise analyses, but not necessary for cluster-wise analysis due to the difference of geometry between format. Once I verify this consistency (and eventually fix any other bugs that would be discovered), I will update the test data and the test script for this PR.

nicholst:
Super!  Thanks so much for keeping up on this!

BryanGuillaume:
@nicholst, the test data and test scripts have been updated and a few remaining bugs have been fixed. Now, all the tests are passing and are covering all 4 types of inputs (i.e. NIfTI, MAT, GIfTI and CIfTI).  I believe this PR can be merged now and any further bugs/improvements should be handled by other PRs. Can I merge it? 

As a side note, the individual tests are also now faster than before, which was intended, but still takes between 2 and 3 minutes each (before some tests could take between 10 to 20 minutes). I was expecting them to be faster than that. It seems the main bottleneck is now the configuration of the tests rather than the test themselves. Here a screenshot of a test job,

Screenshot 2020-06-12 at 15 48 43
 
We can see the slowest calls are from

- `docker pull cmaumet/octave-spm`
- `did=$(sudo docker run -ti -d --rm -v `pwd`:/swe cmaumet/octave-spm)`

and that is where we would need to act to make the tests faster if this was needed in the future.

nicholst:
Thank you @BryanGuillaume !  This looks great.  But I fear I've just discovered something I should have caught immediately.  This formula is wrong!
```
k_Z = swe_invNcdf(0.75) * (boxcox(lambda, k_E) - Q2) / (Q3-Q2))
```

The normalisation has to be "How to scale an IQR standardised value into a SD-standardised value?  You do that by dividing by the IQR of a standard normal... which of course is Q3-Q2 for a standard normal.  So this should be:

```
k_Z = (swe_invNcdf(0.75) - swe_invNcdf(0.25))* (boxcox(lambda, k_E) - Q2) / (Q3-Q2))
```

Right?

I'm so sorry I didn't see this right away a month ago!

BryanGuillaume:
@nicholst, not sure. Please note that we divide by the upper-half IQR `(Q3-Q2)` instead of the IQR `(Q3-Q1)`. Thus, intuitively, we should multiply by `(swe_invNcdf(0.75) - swe_invNcdf(0.5))`, which is equal to `(swe_invNcdf(0.75) - 0 = swe_invNcdf(0.75)`. 

Am I correct?

nicholst:
Ach!  You’re right! Sorry for the panic!

This is absolutely correct as the hIQR!

So... it you’re happy I’m happy to merge.
-- 
Thomas Nichols, PhD

PR #169, opened 2020-05-05
Replace double quotation syntax by simple quotation syntax

BryanGuillaume:
The goal of this PR is to remove any use of the double quotation syntax. There are two reasons for this:

1. The double quotation syntax is not available in old versions of Matlab (older than R2017a)
2. The double quotation syntax yields a string and not a char array like the simple quotation syntax

The difficulty was to find every use in the code as a simple search of the `"` character yielded hundreds of match for it. Nonetheless, I went through all the files in the code and it seems that I caught every double-quote syntax use.

 As it is a very simple modification of the code, I think this PR is ready to be merged.

### DIFF
--- a/swe_getSPM.m
+++ b/swe_getSPM.m
@@ -938,7 +938,7 @@ if ~isMat
                   elseif Ic == 2
                     maxScore = sort(-SwE.WB.minScore);
                   else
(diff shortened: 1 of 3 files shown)


nicholst:
Thank you! Looks good!

PR #170, opened 2020-05-05
[WIP] Fix a bug occurring when an entire chunk of data is masked out

BryanGuillaume:
The goal of this PR is to fix a bug occurring when an entire chunk of data would be masked out. In this scenario, the toolbox would skip the computation of results for this chunk of data but attempt to save them in an empty variable. If this was the first chunk of data, the result variables of this chunk of data would not have been created and, if it was a later chunk of data, the result variables of the previous chunk of data would actually be kept in memory. As the toolbox would attempt to assign these results variables to an empty variable, the toolbox would throw an error. 

With this PR, the toolbox checks now if each chunk of data has been completely masked out or not. If it is the case, it skips the variable assignment, preventing any erroneous assignment.

The changes are minimal and should fix the issue noted above, but I still need to test the PR on a pathological case to see if there exist other side effects that would occur when an entire chunk of data is masked out.

### DIFF
--- a/swe_cp.m
+++ b/swe_cp.m
@@ -824,7 +824,9 @@ if ~isMat
       %-Write beta files
       %----------------------------------------------------------------------
       for iBeta=1:nBeta
-        c(cmask) = beta(iBeta,:);
+        if CrS
+          c(cmask) = beta(iBeta,:);
+        end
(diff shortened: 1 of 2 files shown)


nicholst:
Looks good to merge.

PR #172, opened 2020-06-03
Replace the function 'replace' by the function 'strrep' as it is compatible with older Matlab versions

BryanGuillaume:
The goal of this PR is to replace the relatively recent Matlab function `replace` (introduced in R2016b) by the much older Matlab function `strrep` (introduced before R2006a) in order to allow the use of the toolbox with old Matlab versions.

@nicholst, this PR is straightforward and I think we can merge it straightaway.


nicholst:
Yes! Good call... small changes that can help with backwards compatibility are always appreciated by users.

Merge away!

Issue #173, opened 2020-06-04
Pad short contrast vectors with zeros


nicholst:
Currently an error, "contrast not well specified" is generated if the contrast does not have exactly the same number of rows as the number of predictors.

In SPM, trailing zeros of a contrast can be omitted and are padded automatically.  Adding this feature will help users coming from SPM, especially with designs with lots of nuisance variables.

nicholst:
Closed by #174

PR #174, opened 2020-06-04
Zero pad contrasts

nicholst:
Zero pad short T and F contrast vectors, to match SPM functionality.

Also, provide more nuanced errors when contrasts are malformed.

Addresses #173.


BryanGuillaume:
I confirm that no current test will cover this.  After reviewing the changes, they seem fine to me. Thus, I will directly merge this PR.

PR #175, opened 2020-06-05
Display scaled design matrix

nicholst:
SPM uses some arbitrary heuristics to create a visually appealing design matrix, saved as the variable `xX.nKX`.  While currently, this scaled design matrix is shown when the model is configured (https://github.com/NISOx-BDI/SwE-toolbox/blob/master/swe_run_smodel.m#L888), later when the model is actually examined in Results the raw design matrix `xX.X` is used (https://github.com/NISOx-BDI/SwE-toolbox/blob/master/swe_results_ui.m#L619).

This PR does two things: (1) It changes `swe_results_ui` to use `xX.nKX` and (2) it adds code to `swe_cp` and `swe_cp_WB` to ensure that variable is created.  (In `swe_run_smodel`, `swe_DesRep` is called which creates `xX.nKX` if it doesn't exist.  The edits to `swe_results_ui` were made in reference to the current SPM12 `spm_results_ui` and I took the opportunity to rebase and add explicit reference the graphic handles. (Hopefully that doesn't come back to bite us). 

As a side note, the name `nKX` alludes to `KX` which is the filtered/whitened design matrix; as we always use an identity working covariance matrix KX = X.

### DIFF
--- a/swe_conman.m
+++ b/swe_conman.m
@@ -633,8 +633,8 @@ switch lower(varargin{1})
         %-Picture design matrix
         %------------------------------------------------------------------
         axes(h)
-        if isfield(varargin{2},'nKX') && ~isempty(varargin{2}.X)
-            hDesMtxIm = image((varargin{2}.X+1)*32);
+        if isfield(varargin{2},'nKX') && ~isempty(varargin{2}.nKX)
+            hDesMtxIm = image((varargin{2}.nKX+1)*32);
         else
             hDesMtxIm = image(...
                 (spm_DesMtx('sca',varargin{2}.X,varargin{2}.name)+1)*32);
--- a/swe_cp.m
+++ b/swe_cp.m
@@ -1119,6 +1119,7 @@ end
 % end
 SwE.VM         = VM;                %-Filehandle - Mask
 
+xX.nKX         = spm_DesMtx('sca',xX.X,xX.name); %-Scaled DesMtx for display
 SwE.xX         = xX;                %-design structure
 SwE.xM         = xM;                %-mask structure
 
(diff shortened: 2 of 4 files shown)


nicholst:
@BryanGuillaume again an other small PR that will not have any test coverage... please inspect carefully and merge if you think its ready.

BryanGuillaume:
The changes seem fine to me except for a weird side effect, described later. 

First, I confirm that, in the result displays, the scaled design matrix is now displayed instead of the raw design matrix. For example, I got the following design matrix on a local test:
Screenshot 2020-06-09 at 22 23 28
which is what we want. 

Nevertheless, this nice design matrix representation is not used anymore by the contrast manager:
Screenshot 2020-06-09 at 22 22 45
while this was surprisingly the case before. This weird side effect seems to be due because the PR is adding `nKX` to `SwE.xX`, which cause the raw design matrix to be displayed in place of the scaled version as can be observed in the piece of code:

https://github.com/NISOx-BDI/SwE-toolbox/blob/5f84bdc1025dc452c7a3a943d2f8107d23713cfa/swe_conman.m#L636-L641

Changing line 637 by `hDesMtxIm = image((varargin{2}.nKX+1)*32); `should solve the observed issue. @nicholst, I would change this myself, but I am unsure if I can actually modify a PR which is not from me.

nicholst:
Thanks @BryanGuillaume!  You could have cloned my fork and then lodged a PR against the DisplayScaledDesMtx branch, but, no worries, I've made the edit, now, see: https://github.com/NISOx-BDI/SwE-toolbox/pull/175/commits/cd86ba8cc2baffe653a8fa52d28615799ffb0000

BryanGuillaume:
After this last commit, everything seems to work as intended. Thus, I will merge this PR now.

Issue #176, opened 2020-06-09
Bug due to the fact that variable cmask in swe_cp is a column vector for GIfTI and CIfTI 


BryanGuillaume:
While reviewing PR #175 , I have noted that the variable `cmask` is a row vector for NIfTI, but a column vector for GIfTI and CIfTI. The latter can cause some unwanted effects in some places in the code. For example, the piece of code https://github.com/NISOx-BDI/SwE-toolbox/blob/5f84bdc1025dc452c7a3a943d2f8107d23713cfa/swe_cp.m#L696-L700
works only if `cmask` is a row vector. Indeed, if it is a column vector, the variable `diffVis` becomes a matrix after the loop due to the `|` operation that is occurring between a row and a column vector inside the loop.

As a side note, the problematic piece of code above is not always visited when doing an analysis and was not visited by the tests done so far for GIfTI and CIfTI.  This explains why the bug has not been detected before now.

It seems that forcing `cmask` to be a row vector like for NIfTI analyses would prevent the issue noted above. But it seems important to check if this change would have some other unwanted side effects in some other parts of the code.

nicholst:
OK, thanks @BryanGuillaume! Not sure of the implications of transposing`cmask` right now...

BryanGuillaume:
This issue was solved in PR #177

PR #177, opened 2020-06-13
Ensure the variable cmask is a row vector for CIfTI and GIFTI

BryanGuillaume:
The goal of this PR is to fix issue #176. The fix checks if the variable `cmask` has a single column and, if it is the case, it transposes it.

Some tests on my computer seem to indicate there is no side effect to do so.  The Travis tests should actually cover this change. Thus, if they pass, I think we can assume that it did not break anything.


BryanGuillaume:
I am merging this PR as all the automatic tests and additional tests on my computer have passed

Issue #165, opened 2020-04-11
spm_matlab_version_chk does not seem to be used anymore in SPM while SwE still uses it


BryanGuillaume:
It seems that `spm_matlab_version_chk` is not used anymore in the most recent versions of SPM while it was used in less recent versions (e.g., SPM8). In the SwE toolbox, we use it to save the `SwE.mat` file:
```
%-Save analysis parameters in SPM.mat file
%--------------------------------------------------------------------------
if spm_matlab_version_chk('7') >=0
    save('SPM.mat','SPM','-V6');
else
    save('SPM.mat','SPM');
end

```
In the most recent version of SPM12, the piece of code saving the `SPM.mat` files does not use this type of piece of code anymore, but something like
```
fmt = spm_get_defaults('mat.format');
s = whos('SPM');
if s.bytes > 2147483647, fmt = '-v7.3'; end
save('SPM.mat','SPM', fmt);
```
 and I wonder if we should update the SwE toolbox to follow this saving style for the `SwE.mat` as well or if it is fine to leave it as is.

That needs to be reviewed.

nicholst:
Interesting... so instead of `spm_matlab_version_chk.m` we have `spm_check_version.m`.  But, indeed, if *we* in SwE are only checking the version for I/O, i.e. for saving, I'm inclined to copy what Guillaume is doing, yes.

Is this enough to go on?

nicholst:
While the PR #183 addresses this it will introduce errors with old, un-updated versions of SPM12.

The more robust solution is to create a `swe_check_version` that tests for the existance of`spm_check_version` and uses it if it's available, and otherwise falls back to `spm_matlab_version_chk`.

nicholst:
I was halfway through adding a wrapper `swe_check_version` when I realised that our code *already* depends on `spm_check_version` (the new function) in various places.  Hence there is no reason to maintain backwards compatiblity with `spm_matlab_version_chk`.

PR #183, opened 2020-11-16
Remove dependence on old spm_matlab_version_chk

nicholst:
This addresses #165... `spm_matlab_version_chk` has been long depricated; this PR updates it with the modern function, `spm_check_version`.


nicholst:
This is a trivial change and CI checks out... so merging.

